PR #10, opened 2022-09-21
event update

talhamalik883:


contracts/smart-contract-wallet/SmartWallet.sol
```diff
@@ -68,8 +68,9 @@ contract SmartWallet is
 
     function setOwner(address _newOwner) external mixedAuth {
         require(_newOwner != address(0), "Smart Account:: new Signatory address cannot be zero");
+         address oldOwner = owner;
         owner = _newOwner;
-        emit EOAChanged(address(this),owner,_newOwner);
+        emit EOAChanged(address(this), oldOwner, _newOwner);
```
livingrockrises, line 73: if newOnwer is this way : _newOwner
review name of oldOwner. Should it not be _oldOwner ??

livingrockrises (commented):
check comments

contracts/smart-contract-wallet/SmartWallet.sol
talhamalik883: variable name starting with _ is the convention only needs to be followed for arguments. Variable inside the class or function scope does not needs to be declared with _

livingrockrises (approved):
approved

PR #26, opened 2023-02-13
C4 496 smart account implementation initialization

filmakarov:
Added _disableInitializers() to constructor.

Decided not to add signer != address(0) check in checkSignatures as it will add extra gas for every signature verification.
In general, for actual user wallets, there will never be address(0) as owner, so no danger here.

Btw, it can be interesting to redesign the full proxy deployment and initialization flow according to https://github.com/code-423n4/2023-01-biconomy-findings/blob/main/data/0xSmartContract-G.md#g-02-remove-the-initializer-modifier
it can make new scw deployment cheaper


livingrockrises (approved):
LGTM

PR #28, opened 2023-02-14
175 - Insufficient 1271 signature validation

filmakarov:
Remediation was already in place 

`require(_signer == owner, "INVALID_SIGNATURE");`

for all kinds of signatures.

Added test case to ensure the issue is fixed


livingrockrises (approved):
LGTM

PR #32, opened 2023-02-15
C4 489 prevent relayer from stolen refund

livingrockrises:
removed 21000 + msg.data.length from execTransaction

This would change how we send baseGas which is used as external gas for what can't be accounted internally using gasleft()

earlier it would be : events that's left out of gasleft() + handlePayment gas

now it would be : events that's left out of gasleft() + handlePayment gas + 21000 + data gas

test case restructuring and reviews are left.
this would possibly impact backend API / SDK methods that rely on preparing the base gas for forward transaction final payload. cc @talhamalik883 @AmanRaj1608


test/smart-wallet/refund-estimation.ts
```diff
@@ -313,6 +319,19 @@ describe("Wallet tx gas estimations with and without refunds", function () {
...
+      "execTransaction",
+      [
+        transaction,
+        0, // batchId
+        refundInfo,
+        signature,
+      ]
+    );
+
+    const baseGasReal = txBaseCost(dataEstimate);
```
filmakarov, line 332: will it be used somewhere?

test/smart-wallet/refund-estimation.ts
```diff
@@ -446,18 +469,19 @@ describe("Wallet tx gas estimations with and without refunds", function () {
...
     // safeTx.baseGas = handlePaymentEstimate + 17100 - 9794 + 2360 + 5000; // add 5000 offset for delegate call?;
 
-    safeTx.baseGas = handlePaymentEstimate + 17100 - 7306 + 2360 + 5000; // add 5000 offset for delegate call?;
+    console.log(safeTx);
+
+    safeTx.gasPrice = 156849;
+    safeTx.tokenGasPriceFactor = 1000000;
+
+    safeTx.baseGas =
+      21000 + 5000 + handlePaymentEstimate + 17100 - 7306 + 2360 + 5000; // add 5000 offset for delegate call?;
```
filmakarov, line 478: What is 7306?

livingrockrises: Events + internal call  2360
handle ETH/native payment estimation is 7306
17100 i believe was for state changes offset as per very old notes and calcs

However gas estimation test cases needs a lot of rework.

test/smart-wallet/refund-estimation.ts
livingrockrises: I don't think so. we're trying to get initial gas usage that have to be added in baseGas field of fee refund
initial gas is 21000 + depends on 1 and 0 bytes in msg.data

see now there is a circular dependency cause if you want to use right msg.data you need a signature first, and for this signature you'd need accurate baseGas. but the whole point of estimation is to get right baseGas so you can sign it.

filmakarov (approved):

PR #34, opened 2023-02-16
C4 154 Remove Receive() function

filmakarov:
* Removed receive() from proxy
* Can't remove from implementation as it has no payable fallback
* https://github.com/safe-global/safe-contracts/blob/main/contracts/common/EtherPaymentFallback.sol 
* Added event `SmartAccountReceivedNativeToken` to track incoming native token payments

### DIFF
--- a/contracts/smart-contract-wallet/Proxy.sol
+++ b/contracts/smart-contract-wallet/Proxy.sol
@@ -33,8 +33,4 @@ contract Proxy {
         }
     }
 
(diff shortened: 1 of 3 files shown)


livingrockrises (commented):
looks good to me

filmakarov:
> probably add a test case that verifies event added is emitted when you send native tokens to smart account address.

done

PR #35, opened 2023-02-16
BSDK-362 Enable value transfer using executeBatch

AmanRaj1608:
- pass array of values in executeBatch function and remove hardcoded 0 in _call() function. use values[i] instead.

- rename execute method name which conflicts the internal execute method in Executor.sol

- while doing this, also remove onlyOwner modifier on execute() and executeBatch() functions(search 495 - execute() and executeBatch() functions in SmartAccount.sol from the EntryPoint will fail ) in mid risk page

- also combining this here: #313 making sure above functions and any other function that can have reentrancy uses non-reentrant modifier.

### DIFF
--- a/contracts/smart-contract-wallet/SmartAccount.sol
+++ b/contracts/smart-contract-wallet/SmartAccount.sol
@@ -463,16 +463,21 @@ contract SmartAccount is
         SafeERC20.safeTransfer(tokenContract, dest, amount);
     }
 
-    function execute(address dest, uint value, bytes calldata func) external onlyOwner{
+    function executeCall(address dest, uint256 value, bytes calldata func) external {
         _requireFromEntryPointOrOwner();
         _call(dest, value, func);
     }
 
-    function executeBatch(address[] calldata dest, bytes[] calldata func) external onlyOwner{
+    function executeBatchCall(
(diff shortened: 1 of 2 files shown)


contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -463,16 +463,25 @@ contract SmartAccount is
...
+    function executeBatchCall(
+        address[] calldata dest,
+        uint256[] calldata value,
+        bytes[] calldata func
+    ) external nonReentrant {
         _requireFromEntryPointOrOwner();
-        require(dest.length == func.length, "wrong array lengths");
-        for (uint i = 0; i < dest.length;) {
-            _call(dest[i], 0, func[i]);
+        require(dest.length == value.length, "wrong array lengths");
```
filmakarov, line 481: should we explicitly revert when lengths are 0?

AmanRaj1608: resolved

AmanRaj1608: have removed and added a test

livingrockrises (approved):
LGTM

PR #36, opened 2023-02-16
functions with mixedAuth modifiers should have public visibility

AmanRaj1608:
Ownable2Step transfer is not required. As even in the case where user accidentally transfers ownership to a new address which user doesn’t own, he can still use social recovery module to change the owner back.

Changing functions with mixedAuth modifiers to public to be called from both inside and outside the contract via modules.

### DIFF
--- a/contracts/smart-contract-wallet/SmartAccount.sol
+++ b/contracts/smart-contract-wallet/SmartAccount.sol
@@ -121,7 +121,7 @@ contract SmartAccount is
 
     // Setters
 
-    function setOwner(address _newOwner) external mixedAuth {
+    function setOwner(address _newOwner) public mixedAuth {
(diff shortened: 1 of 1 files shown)


filmakarov (approved):

livingrockrises:
do add test cases for setOwner
via owner and via address(this) [ can be a module or from entry point transaction ] 
that checks gull mixedAuth. 

negative test case for the above when via address(this) should not work if some test method in MaliciousAccount is marked external.

AmanRaj1608:
This is more related to #39 

In that PR have removed the `mixedAuth` from `updateImplementation` and have tested both methods.

livingrockrises (approved):
looks good

PR #38, opened 2023-02-20
making proxy and wallet factory lighter 

livingrockrises:
Making proxy and wallet factory lighter  ( commit for review & chat + dev notes )

couple of things to discuss here which I put in dev notes as well (final code will be cleaned up)

**Proxy.sol**

- removed EIP1967 storage slot
- commented out singleton to be at 0th storage slot (gnosis uses this way to fetch the target)
- storing singleton (base implementation / accountLogic / singleton) at storage slot given by address() opcode [ maybe we should pick one name from above three and use through out ]
- removed receive event as it doesn't make sense here 
- added accountLogic read method to query the implementation address on proxy 

** SmartAccountFactory.sol **

- added read method accountCreationCode to get the creation code of Proxy
- changed unit -> unit256
- isAccountExist is commented out for now it's unnecessary state management unless we absolutely need it
- commented out VERSION as versions can conflict with smart account implementation contract versioning [ we can emit an event under initialise to do version tracking ] 
- commented event SmartAccountCreated. added AccountCreation (which may also emit owner and index. but above event inside init can do that as well)

open for discussion (comparison with Safe factory and others)

- need of constructor
- versioning plans if constructor args changes
- if constructor is removed _implementation address can be part of function args. [goes well with not keeping version variable in factory]
- todo : rename wallet to account
- possibility of sending initialiser also as another deploy function (create2) but that would require another read method too to get counterfactual address

### DIFF
--- a/contracts/smart-contract-wallet/Proxy.sol
+++ b/contracts/smart-contract-wallet/Proxy.sol
@@ -5,17 +5,21 @@ pragma solidity 0.8.12;
  * @title Proxy // This is the user's wallet
  * @notice Basic proxy that delegates all calls to a fixed implementing contract.
  */
+ interface IProxy {
+    function accountLogic() external view returns (address);
+}
+
 contract Proxy {
+    // no fixed address(0) storage slot
+    // address internal singleton;
 
-    /* This is the keccak-256 hash of "biconomy.scw.proxy.implementation" subtracted by 1, and is validated in the constructor */
-    bytes32 internal constant _IMPLEMENTATION_SLOT = 0x37722d148fb373b961a84120b6c8d209709b45377878a466db32bbc40d95af26;
-
-    event Received(uint indexed value, address indexed sender, bytes data);
+    // uint256[1] private ______gap;
 
     constructor(address _implementation) {
-         assert(_IMPLEMENTATION_SLOT == bytes32(uint256(keccak256("biconomy.scw.proxy.implementation")) - 1));
+         require(_implementation != address(0), "Invalid implementation address");
+         // solhint-disable-next-line no-inline-assembly
          assembly {
-             sstore(_IMPLEMENTATION_SLOT,_implementation) 
+             sstore(address(),_implementation) 
          }
     }
 
@@ -23,7 +27,12 @@ contract Proxy {
         address target;
         // solhint-disable-next-line no-inline-assembly
         assembly {
-            target := sload(_IMPLEMENTATION_SLOT)
+            target := sload(address())
+            // 0xee9118f7 == keccak("accountLogic()"). The value is right padded to 32-bytes with 0s
+            if eq(calldataload(0), 0xee9118f700000000000000000000000000000000000000000000000000000000) {
+                mstore(0, target)
+                return(0, 0x20)
+            }
             calldatacopy(0, 0, calldatasize())
             let result := delegatecall(gas(), target, 0, calldatasize(), 0, 0)
             returndatacopy(0, 0, returndatasize())
--- a/contracts/smart-contract-wallet/SmartAccountFactory.sol
+++ b/contracts/smart-contract-wallet/SmartAccountFactory.sol
@@ -4,58 +4,85 @@ pragma solidity 0.8.12;
 import "./Proxy.sol";
 import "./BaseSmartAccount.sol"; 
 
+// @todo
+// cleanup comments and review notes below
+
 contract SmartAccountFactory {
-    address immutable public _defaultImpl;
+    address immutable public _implementation;
     address immutable public _defaultFallbackHandler; 
+    // ^^ this means if defaultImpl or defaultFallbackHandler changes then we'd have to deploy new factory.
+    // (as per currrent versioning) defaultImpl changes = version update in impl and proxy both 
+    // defaultFallbackHandler changes = version update only in factory?!
 
-    // EOA + Version tracking
-    string public constant VERSION = "1.0.4";
+    // should be needed to emit from accountLogic : SmartAccountInitialized
+    // string public constant VERSION = "1.0.4";
 
-    //states : registry
-    // review need and impact of this update wallet -> account
-    mapping (address => bool) public isAccountExist;
+    // Review event
+    // event SmartAccountCreated(address indexed _proxy, address indexed _implementation, address indexed _owner, string version, uint256 _index);
 
-    constructor(address _baseImpl, address _handler) {
-        require(_baseImpl != address(0), "base wallet address 0");
-        _defaultImpl = _baseImpl;
-        require(_handler != address(0), "default fallback handler 0");
+     event AccountCreation(address indexed account, address indexed accountLogic);
+
+    // not to check if address is a contract but if it's deployed from this proxy
+    // mapping (address => bool) public isAccountExist;
+
+    constructor(address _singleton, address _handler) {
+        require(_singleton != address(0), "invalid singleton address");
+        _implementation = _singleton;
+        require(_handler != address(0), "invalid fallback handler");
         _defaultFallbackHandler = _handler;
     }
 
-    // Review event
-    event SmartAccountCreated(address indexed _proxy, address indexed _implementation, address indexed _owner, string version, uint256 _index);
+    /// @dev Allows to retrieve the creation code used for the Proxy deployment.
+    function accountCreationCode() public pure returns (bytes memory) {
+        return type(Proxy).creationCode;
+    }
 
     /**
-     * @notice Deploys wallet using create2 and points it to _defaultImpl
+     * @notice Deploys wallet using create2 and points it to _implementation
      * @param _owner EOA signatory of the wallet
      * @param _index extra salt that allows to deploy more wallets if needed for same EOA (default 0)
      */
-    function deployCounterFactualWallet(address _owner, uint _index) public returns(address proxy){
+    function deployCounterFactualWallet(address _owner, uint256 _index) public returns(address proxy){
+        // check optimisation scope in creating salt...
         bytes32 salt = keccak256(abi.encodePacked(_owner, address(uint160(_index))));
-        bytes memory deploymentData = abi.encodePacked(type(Proxy).creationCode, uint(uint160(_defaultImpl)));
+
+        bytes memory deploymentData = abi.encodePacked(type(Proxy).creationCode, uint256(uint160(_implementation)));
+
         // solhint-disable-next-line no-inline-assembly
         assembly {
             proxy := create2(0x0, add(0x20, deploymentData), mload(deploymentData), salt)
         }
         require(address(proxy) != address(0), "Create2 call failed");
+
         // EOA + Version tracking
-        emit SmartAccountCreated(proxy,_defaultImpl,_owner, VERSION, _index);
+        // emit SmartAccountCreated(proxy,_defaultImpl,_owner, VERSION, _index);
+
+        // you can pass initializer data but then that it also needs to be part of salt
+        // init method name subject to change
         BaseSmartAccount(proxy).init(_owner, _defaultFallbackHandler);
-        isAccountExist[proxy] = true;
+
+        // isAccountExist[proxy] = true;
+        emit AccountCreation(proxy, _implementation);
     }
 
     /**
-     * @notice Deploys wallet using create and points it to _defaultImpl
+     * @notice Deploys wallet using create and points it to _implementation
      * @param _owner EOA signatory of the wallet
     */ 
     function deployWallet(address _owner) public returns(address proxy){ 
-        bytes memory deploymentData = abi.encodePacked(type(Proxy).creationCode, uint(uint160(_defaultImpl)));
+        bytes memory deploymentData = abi.encodePacked(type(Proxy).creationCode, uint256(uint160(_implementation)));
         // solhint-disable-next-line no-inline-assembly
         assembly {
             proxy := create(0x0, add(0x20, deploymentData), mload(deploymentData))
         }
+        require(address(proxy) != address(0), "Create call failed");
+
+        // you can pass initializer data but then that it also needs to be part of salt
+        // init method name subject to change
         BaseSmartAccount(proxy).init(_owner, _defaultFallbackHandler);
-        isAccountExist[proxy] = true;
+        // isAccountExist[proxy] = true;
+
+        emit AccountCreation(proxy, _implementation);
     }
 
     /**
@@ -63,8 +90,8 @@ contract SmartAccountFactory {
      * @param _owner EOA signatory of the wallet
      * @param _index extra salt that allows to deploy more wallets if needed for same EOA (default 0)
     */
-    function getAddressForCounterfactualWallet(address _owner, uint _index) external view returns (address _wallet) {
-       bytes memory code = abi.encodePacked(type(Proxy).creationCode, uint(uint160(_defaultImpl)));
+    function getAddressForCounterfactualWallet(address _owner, uint256 _index) external view returns (address _wallet) {
+       bytes memory code = abi.encodePacked(type(Proxy).creationCode, uint256(uint160(_implementation)));
        bytes32 salt = keccak256(abi.encodePacked(_owner, address(uint160(_index))));
        bytes32 hash = keccak256(abi.encodePacked(bytes1(0xff), address(this), salt, keccak256(code)));
         _wallet = address(uint160(uint(hash)));
--- a/test/smart-wallet/gasEstimations.ts
+++ b/test/smart-wallet/gasEstimations.ts
@@ -165,8 +165,8 @@ describe("Wallet tx gas estimations with and without refunds", function () {
     console.log("deploying new wallet..expected address: ", expected);
 
     await expect(walletFactory.deployCounterFactualWallet(owner, indexForSalt))
-      .to.emit(walletFactory, "SmartAccountCreated")
-      .withArgs(expected, baseImpl.address, owner, "1.0.4", indexForSalt);
+      .to.emit(walletFactory, "AccountCreation")
+      .withArgs(expected, baseImpl.address);
 
     userSCW = await ethers.getContractAt(
       "contracts/smart-contract-wallet/SmartAccount.sol:SmartAccount",
--- a/test/smart-wallet/proxy-deploy.ts
+++ b/test/smart-wallet/proxy-deploy.ts
@@ -41,8 +41,15 @@ describe("Wallet Deployment", function () {
     );
     console.log("deploying new wallet..expected address: ", expected);
 
-    await expect(walletFactory.deployCounterFactualWallet(owner, indexForSalt))
-      .to.emit(walletFactory, "SmartAccountCreated")
-      .withArgs(expected, baseImpl.address, owner, "1.0.4", 0);
+    const tx = await walletFactory.deployCounterFactualWallet(
+      owner,
+      indexForSalt
+    );
+    const receipt = await tx.wait();
+    console.log("smart account deployment gas ", receipt.gasUsed.toNumber());
+
+    /* await expect(walletFactory.deployCounterFactualWallet(owner, indexForSalt))
+      .to.emit(walletFactory, "AccountCreation")
+      .withArgs(expected, baseImpl.address); */
   });
 });
--- a/test/smart-wallet/refund-estimation.ts
+++ b/test/smart-wallet/refund-estimation.ts
@@ -174,8 +174,8 @@ describe("Wallet tx gas estimations with and without refunds", function () {
     console.log("deploying new wallet..expected address: ", expected);
 
     await expect(walletFactory.deployCounterFactualWallet(owner, indexForSalt))
-      .to.emit(walletFactory, "SmartAccountCreated")
-      .withArgs(expected, baseImpl.address, owner, "1.0.4", indexForSalt);
+      .to.emit(walletFactory, "AccountCreation")
+      .withArgs(expected, baseImpl.address);
 
     userSCW = await ethers.getContractAt(
       "contracts/smart-contract-wallet/SmartAccount.sol:SmartAccount",
--- a/test/smart-wallet/testGroup1.ts
+++ b/test/smart-wallet/testGroup1.ts
@@ -114,8 +114,8 @@ describe("Base Wallet Functionality", function () {
     console.log("deploying new wallet..expected address: ", expected);
 
     await expect(walletFactory.deployCounterFactualWallet(owner, indexForSalt))
-      .to.emit(walletFactory, "SmartAccountCreated")
-      .withArgs(expected, baseImpl.address, owner, VERSION, indexForSalt);
+      .to.emit(walletFactory, "AccountCreation")
+      .withArgs(expected, baseImpl.address);
 
     userSCW = await ethers.getContractAt(
       "contracts/smart-contract-wallet/SmartAccount.sol:SmartAccount",
@@ -143,9 +143,8 @@ describe("Base Wallet Functionality", function () {
     });
 
     await expect(tx)
-      .to.emit(userSCW, 'SmartAccountReceivedNativeToken')
+      .to.emit(userSCW, "SmartAccountReceivedNativeToken")
       .withArgs(bob, ethers.utils.parseEther("5"));
-
   });
 
   // Transactions
@@ -242,11 +241,9 @@ describe("Base Wallet Functionality", function () {
     let signature = "0x";
     signature += data.slice(2);
     await expect(
-      userSCW.connect(accounts[0]).execTransaction(
-        transaction,
-        refundInfo,
-        signature
-      )
+      userSCW
+        .connect(accounts[0])
+        .execTransaction(transaction, refundInfo, signature)
     ).to.emit(userSCW, "ExecutionSuccess");
 
     await expect(
@@ -310,11 +307,9 @@ describe("Base Wallet Functionality", function () {
     signature += data.slice(2);
 
     await expect(
-      userSCW.connect(accounts[0]).execTransaction(
-        transaction,
-        refundInfo,
-        signature
-      )
+      userSCW
+        .connect(accounts[0])
+        .execTransaction(transaction, refundInfo, signature)
     ).to.emit(userSCW, "ExecutionSuccess");
 
     expect(await token.balanceOf(charlie)).to.equal(
@@ -364,17 +359,15 @@ describe("Base Wallet Functionality", function () {
     let signature = "0x";
     signature += data.slice(2);
     await expect(
-      userSCW.connect(accounts[0]).execTransaction(
-        transaction,
-        refundInfo,
-        signature
-      )
+      userSCW
+        .connect(accounts[0])
+        .execTransaction(transaction, refundInfo, signature)
     ).to.be.reverted;
 
     expect(await token.balanceOf(charlie)).to.equal(
       ethers.utils.parseEther("0")
     );
-  }); 
+  });
 
   it("Can not execute txn with the same nonce twice", async function () {
     await token
@@ -417,11 +410,9 @@ describe("Base Wallet Functionality", function () {
     signature += data.slice(2);
 
     await expect(
-      userSCW.connect(accounts[0]).execTransaction(
-        transaction,
-        refundInfo,
-        signature
-      )
+      userSCW
+        .connect(accounts[0])
+        .execTransaction(transaction, refundInfo, signature)
     ).to.emit(userSCW, "ExecutionSuccess");
 
     expect(await token.balanceOf(charlie)).to.equal(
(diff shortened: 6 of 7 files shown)


contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -176,7 +181,7 @@ contract SmartAccount is
     // init
     // Initialize / Setup
     // Used to setup
-    function init(address _owner, address _handler) public override initializer {
```
AmanRaj1608, line 179: Can add OZ __ReentrancyGuard_init(); in init() method.

contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -176,7 +181,7 @@ contract SmartAccount is
     // init
     // Initialize / Setup
     // Used to setup
-    function init(address _owner, address _handler) public override initializer { 
+    function init(address _owner, address _handler) public override { 
         require(owner == address(0), "Already initialized");
         require(_owner != address(0),"Invalid owner");
         require(_handler != address(0), "Invalid Fallback Handler");
```
AmanRaj1608, line 187: we can move the _handler check to `setFallbackHandler`, as `setFallbackHandler` still doesn’t have any 0 address check

test/proxy-upgrades/test-upgrade.ts
```diff
@@ -0,0 +1,267 @@
...
+    const MultiSend = await ethers.getContractFactory("MultiSend");
+    multiSend = await MultiSend.deploy();
+    console.log("Multisend helper contract deployed at: ", multiSend.address);
+
+    console.log("mint tokens to owner address..");
+    await token.mint(owner, ethers.utils.parseEther("1000000"));
+  });
+
+  // describe("Wallet initialization", function () {
+  it("Should set the correct states on proxy", async function () {
```
AmanRaj1608, line 115: Could add a test where we upgrade while decrementing the _gap and introduce some new state and verify everything still behaves expected.

contracts/smart-contract-wallet/base/FallbackManager.sol
```diff
@@ -50,4 +50,6 @@ contract FallbackManager is SelfAuthorized {
             return(0, returndatasize())
         }
     }
+
+    uint256[24] private __gap;
```
AmanRaj1608, line 54: shouldn't we name this with different variable name rather than __gap which is same in ModuleManager too

contracts/smart-contract-wallet/base/ModuleManager.sol
```diff
@@ -130,4 +130,6 @@ contract ModuleManager is SelfAuthorized, Executor {
             mstore(array, moduleCount)
         }
     }
+
+    uint256[24] private __gap;
```
AmanRaj1608, line 134: this

test/proxy-upgrades/test-upgrade.ts
livingrockrises: there is test case for this. state-upgrade-base.ts

contracts/smart-contract-wallet/base/FallbackManager.sol
livingrockrises: I think it doesn't matter since it's private.

livingrockrises: If we want specific names for readability we can name __moduleManagerGap and __fallbackManagerGap

contracts/smart-contract-wallet/base/ModuleManager.sol
livingrockrises: ?

contracts/smart-contract-wallet/SmartAccount.sol
livingrockrises: let me check

livingrockrises:
Also think about which methods 
- we must make virtual so it's easy to override by inheriting base
- are likely to introduce new params but have to follow interface / abstract contracts

contracts/smart-contract-wallet/SmartAccount.sol
livingrockrises: done as part of code security: check this commit https://github.com/bcnmy/scw-contracts/commit/285958386f1c7baac0a5f72b48d5f9b78f71f573

livingrockrises:
need to sit down for implementing Reentrancy Guard again and make a list of all functions and it's parent functions where this guard should be applied. @AmanRaj1608 @filmakarov

contracts/smart-contract-wallet/SmartAccount.sol
AmanRaj1608: cool

livingrockrises: i removed it in that commit..

livingrockrises:
made discussed change in the SmartAccountFactory regarding emitting events and making independent of handler and implementation, hence passing both as params and hence also making handler part of the salt.

filmakarov:
> base implementation / accountLogic / singleton - maybe we should pick one name from above three and use through out

I vote for base implementation

filmakarov:
> if constructor is removed _implementation address can be part of function args. [goes well with not keeping version variable in factory]

Do you mean, that we
a) remove constructor from Proxy
b) use _implementation from `deployCounterFactualWallet` function args to make the first init call, in course of which we store _implementation for this proxy?

filmakarov:
> possibility of sending initialiser also as another deploy function (create2) but that would require another read method too to get counterfactual address

I'm not sure, I get what you mean here. Can you please explain?

filmakarov:
I have an idea, what if we add function like 

```
function initSetImplementation(address _implementation) public {
    assembly {
        if (not(iszero(sload(address())) { revert (0,0) }
        if (iszero(_implementation)) { revert (0,0) }
        sstore(address(),_implementation)
    }
}
```

and we call it in the same deployCounterFactualWallet transaction immediately after deploying proxy.
This will allow us to make creation code independent of implementation and we solve the problem with deploying same wallet address for all chains even if the implementation was changed.

livingrockrises:
> > if constructor is removed _implementation address can be part of function args. [goes well with not keeping version variable in factory]
> 
> Do you mean, that we a) remove constructor from Proxy b) use _implementation from `deployCounterFactualWallet` function args to make the first init call, in course of which we store _implementation for this proxy?
> 
> Also, what is the reason to remove the constructor? making new proxy address independent of implementation? that would be cool

no no I meant to remove constructor from SmartAccountFactory. check this commit https://github.com/bcnmy/scw-contracts/commit/44aeebb3a876818309f40bd51876f1c8d37e7b92

livingrockrises:
> I have an idea, what if we add function like
> 
> ```
> function initSetImplementation(address _implementation) public {
>     assembly {
>         if (not(iszero(sload(address())) { revert (0,0) }
>         if (iszero(_implementation)) { revert (0,0) }
>         sstore(address(),_implementation)
>     }
> }
> ```
> 
> and we call it in the same deployCounterFactualWallet transaction immediately after deploying proxy. This will allow us to make creation code independent of implementation and we solve the problem with deploying same wallet address for all chains even if the implementation was changed.

I was thinking about this in bed. 
for this we'd have to deploy proxy with implementation pointing somewhere, which may be zero address as per code logic above. If we do this and then go about doing initSetImplementation then it wouldn't point to correct implementation in the first place, no? 
that brings back to changing from UUPS to must have update logic in the proxy. 

also above needs formal verification as implementation can be front runned and when we make it at least part of the salt it creates dependency again on the address.

livingrockrises:
> > possibility of sending initialiser also as another deploy function (create2) but that would require another read method too to get counterfactual address
> 
> I'm not sure, I get what you mean here. Can you please explain?

oh I just meant to have another method like 

function deployCounterFactualWallet(address _implementation, address _handler, address _owner, uint256 _index) public returns(address proxy){

can have a cousin

function deployCounterFactualWalletWithInitializer(address _implementation, bytes memory initializer, uint256 _index) public returns(address proxy){

livingrockrises:
> > base implementation / accountLogic / singleton - maybe we should pick one name from above three and use through out
> 
> I vote for base implementation. Or just implementation.

cool. let me review where all it needs to be modified.

filmakarov:
> no no I meant to remove constructor from SmartAccountFactory. check this commit [44aeebb](https://github.com/bcnmy/scw-contracts/commit/44aeebb3a876818309f40bd51876f1c8d37e7b92)

I think factory looks great w/o constructor

filmakarov:
>  If we do this and then go about doing initSetImplementation then it wouldn't point to correct implementation in the first place, no? 

What I suggest is to add this `initSetImplementation` to Proxy itself. So it doesn't move us from UUPS coz it is just one-time function, it's like constructor but can be called after deployment.

> also above needs formal verification as implementation can be front runned and when we make it at least part of the salt it creates dependency again on the address.

Yes, that brings the problem with front-running the (deployment+initSetImplementation) txn. 
We can introduce verification, like check where does the proxy points ( address() slot ) and if it points to suspicious implementation we can warn user not to send funds there. 
However, it doesn't protect users from just cybersquatting addresses. like just frontrun and point proxy to 0xeeeee so it's unusable.

So, I see 2 options:

1) just make implementation part of creation code (as it is now) so we are protected from front-running, but we can have issues with keeping same address for user on all the chains.

2) we can introduce additional checks on `initSetImplementation`
For example it can not just store implementation to address() slot , but it can delegatecall to implementation with "updateImplementation" , so we at least know there is such a method. it's not 100% protection however.

filmakarov:
> function deployCounterFactualWalletWithInitializer(address _implementation, bytes memory initializer, uint256 _index) public returns(address proxy){

I think that's a good idea. 
if we use initializer data the way Safe does 
`call(gas(), proxy, 0, add(initializer, 0x20), mload(initializer), 0, 0)`
and not use it for deploymentData, we won't need other read method I think

filmakarov:
> > need to sit down for implementing Reentrancy Guard again and make a list of all functions and it's parent functions where this guard should be applied. @AmanRaj1608 @filmakarov
> let's tackle this as well asap

Let me  study Safe's Guards. They are not very intuitive

contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -71,7 +67,6 @@ contract SmartAccount is
         owner = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
```
tomarsachin2271, line 67: Are we using this variable in our contract?

mapping(bytes32 => uint256) public signedMessages;

contracts/smart-contract-wallet/SmartAccountFactory.sol
```diff
@@ -4,70 +4,91 @@ pragma solidity 0.8.12;
...
         }
-        BaseSmartAccount(proxy).init(_owner, _defaultFallbackHandler);
-        isAccountExist[proxy] = true;
+        require(address(proxy) != address(0), "Create call failed");
+
+        // you can pass initializer data but then that it also needs to be part of salt
+        // init method name subject to change
+        BaseSmartAccount(proxy).init(_owner, _handler);
+
+        emit AccountCreation(proxy, _implementation, _handler, _owner, 0);
```
tomarsachin2271, line 74: We should emit different event here. 
As indexer indexing these events will get wrong information if we emit same AccountCreation event with index value as 0, even though in that method index is not used at all during creation.

contracts/smart-contract-wallet/SmartAccount.sol
livingrockrises: yep, it's used to mark approved messages as part ofEIP1271 flow

contracts/smart-contract-wallet/Proxy.sol
```diff
@@ -5,25 +5,32 @@ pragma solidity 0.8.12;
  * @title Proxy // This is the user's wallet
  * @notice Basic proxy that delegates all calls to a fixed implementing contract.
  */
-contract Proxy {
-
-    /* This is the keccak-256 hash of "biconomy.scw.proxy.implementation" subtracted by 1, and is validated in the constructor */
-    bytes32 internal constant _IMPLEMENTATION_SLOT = 0x37722d148fb373b961a84120b6c8d209709b45377878a466db32bbc40d95af26;
+ interface IProxy {
```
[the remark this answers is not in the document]
livingrockrises: using this interface someone can call method on proxy to read the implementation. how the method responds is written in assembly by checking calldata in fallback method

contracts/smart-contract-wallet/SmartAccountFactory.sol
livingrockrises: we can emit a different one when CREATE is used

livingrockrises:
> > function deployCounterFactualWalletWithInitializer(address _implementation, bytes memory initializer, uint256 _index) public returns(address proxy){
> 
> I think that's a good idea. if we use initializer data the way Safe does `call(gas(), proxy, 0, add(initializer, 0x20), mload(initializer), 0, 0)` and not use it for deploymentData, we won't need other read method I think


> > If we do this and then go about doing initSetImplementation then it wouldn't point to correct implementation in the first place, no?
> 
> What I suggest is to add this `initSetImplementation` to Proxy itself. So it doesn't move us from UUPS coz it is just one-time function, it's like constructor but can be called after deployment.
> 
> > also above needs formal verification as implementation can be front runned and when we make it at least part of the salt it creates dependency again on the address.
> 
> Yes, that brings the problem with front-running the (deployment+initSetImplementation) txn. We can introduce verification, like check where does the proxy points ( address() slot ) and if it points to suspicious implementation we can warn user not to send funds there. However, it doesn't protect users from just cybersquatting addresses. like just frontrun and point proxy to 0xeeeee so it's unusable.
> 
> So, I see 2 options:
> 
> 1. just make implementation part of creation code (as it is now) so we are protected from front-running, but we can have issues with keeping same address for user on all the chains.
> 2. we can introduce additional checks on `initSetImplementation`
>    For example it can not just store implementation to address() slot , but it can delegatecall to implementation with "updateImplementation" , so we at least know there is such a method. it's not 100% protection however.

I see with initSetImplementation in proxy but like you said it can open up unknown attack vectors. 
with the second point also, since deployCounterFactual is public method we can't do anything access control based from here, probably meaning different implementation can be passed and wallet address won't show change. 

1st can be fine if we know which implementation, which handler, which owner and which index to use while deploying. and once done update implementation and (possibly) set new handler.

livingrockrises:
> > function deployCounterFactualWalletWithInitializer(address _implementation, bytes memory initializer, uint256 _index) public returns(address proxy){
> 
> I think that's a good idea. if we use initializer data the way Safe does `call(gas(), proxy, 0, add(initializer, 0x20), mload(initializer), 0, 0)` and not use it for deploymentData, we won't need other read method I think

I now think it would be the same thing without any benefit. we either pass handler and owner or pass initialiser data and in both case they have to be part of the salt. in case of initializer we'd have to remember that to create a salt, in current way we have to remember handler to regenerate same proxy address. 

only benefit maybe let's say we know handler is wrong and we just needed it for proxy address but want to initialise with right handler afterwards then initialiser data can be sent 0 to save unnecessary init call on proxy.

livingrockrises:
> > > function deployCounterFactualWalletWithInitializer(address _implementation, bytes memory initializer, uint256 _index) public returns(address proxy){
> > 
> > 
> > I think that's a good idea. if we use initializer data the way Safe does `call(gas(), proxy, 0, add(initializer, 0x20), mload(initializer), 0, 0)` and not use it for deploymentData, we won't need other read method I think
> 
> I now think it would be the same thing without any benefit. we either pass handler and owner or pass initialiser data and in both case they have to be part of the salt. in case of initializer we'd have to remember that to create a salt, in current way we have to remember handler to regenerate same proxy address.
> 
> only benefit maybe let's say we know handler is wrong and we just needed it for proxy address but want to initialise with right handler afterwards then initialiser data can be sent 0 to save unnecessary init call on proxy.

but no, in above case also i'd have to send initialiser to get the right salt. 
so besides owner and index there are this values,

_implementationUsedForDeployment
-implmentationDesired
_handlerUsedForDeployment
_handlerDesired

Also keeping in mind that we can't leave proxy uninitialized

filmakarov:
> What do you mean by interace for future improvements? in the context of keeping Guard manager like gnosis?

I mean their checkTransaction is now

```
function checkTransaction(
        address,
        uint256,
        bytes memory,
        Enum.Operation,
        uint256,
        uint256,
        uint256,
        address,
        // solhint-disable-next-line no-unused-vars
        address payable,
        bytes memory,
        address
    ) external override {
        GuardValue storage guard = getGuard();
        require(!guard.active, "Reentrancy detected");
        guard.active = true;
    }
```

they do not use any of the tx info for the verification.
but since they manage guards thru GuardManager, they will be able to easily migrate to a new Guard implementation with more sophisticated checking algorithm, that may use some of the Tx data. And all of that without migrating to the new implementation. 
It also allows in theory for different implementations to even have different Guards. 

Also I think that this architecture allows Guards to be not only re-entrancy guards, but also other kinds of guards that allow or disallow certain kinds of operations.
I think this architecture can be used to control some txns from modules. or we can take inspiration from Gelato's tx whitelisting concept (from their Safe automation module). To whitelist certain txns for certain modules. This can even be incorporated with enableModule flow.

filmakarov:
> I now think it would be the same thing without any benefit. we either pass handler and owner or pass initialiser data and in both case they have to be part of the salt. in case of initializer we'd have to remember that to create a salt, in current way we have to remember handler to regenerate same proxy address.

Yep, I agree here. What I'am trying to say, that passing initializer data is a more universal approach. If we use initializer, even if `init` args or even name will change in the new implementation, we won't need to deploy new factory.

owner and handler will be the part of initializer  (encoded args of the init function) and the salt will depend on initializer and thus on owner and handler

filmakarov:
>_implementationUsedForDeployment
-implmentationDesired
_handlerUsedForDeployment
_handlerDesired
> Also keeping in mind that we can't leave proxy uninitialized

yes, it's a good idea, we can send implmentationDesired + _handlerDesired to the contract use _handlerDesired in the init call to proxy, since we don't have a function to update handler.

livingrockrises:
> > What do you mean by interace for future improvements? in the context of keeping Guard manager like gnosis?
> 
> I mean their checkTransaction is now
> 
> ```
> function checkTransaction(
>         address,
>         uint256,
>         bytes memory,
>         Enum.Operation,
>         uint256,
>         uint256,
>         uint256,
>         address,
>         // solhint-disable-next-line no-unused-vars
>         address payable,
>         bytes memory,
>         address
>     ) external override {
>         GuardValue storage guard = getGuard();
>         require(!guard.active, "Reentrancy detected");
>         guard.active = true;
>     }
> ```
> 
> they do not use any of the tx info for the verification. but since they manage guards thru GuardManager, they will be able to easily migrate to a new Guard implementation with more sophisticated checking algorithm, that may use some of the Tx data. And all of that without migrating to the new implementation. It also allows in theory for different implementations to even have different Guards.
> 
> Also I think that this architecture allows Guards to be not only re-entrancy guards, but also other kinds of guards that allow or disallow certain kinds of operations. I think this architecture can be used to control some txns from modules. or we can take inspiration from Gelato's tx whitelisting concept (from their Safe automation module). To whitelist certain txns for certain modules. This can even be incorporated with enableModule flow.

Got it. Interesting!
I will add a test case to upgrade to an implementation which supports basic functional guards, even if we don't support it in the first implementation we're confident.

livingrockrises:
> > I now think it would be the same thing without any benefit. we either pass handler and owner or pass initialiser data and in both case they have to be part of the salt. in case of initializer we'd have to remember that to create a salt, in current way we have to remember handler to regenerate same proxy address.
> 
> Yep, I agree here. What I'am trying to say, that passing initializer data is a more universal approach. If we use initializer, even if `init` args or even name will change in the new implementation, we won't need to deploy new factory.
> 
> owner and handler will be the part of initializer (encoded args of the init function) and the salt will depend on initializer and thus on owner and handler

I see, good idea. I will make this change and update the test cases.

livingrockrises:
> > _implementationUsedForDeployment
> > -implmentationDesired
> > _handlerUsedForDeployment
> > _handlerDesired
> > Also keeping in mind that we can't leave proxy uninitialized
> 
> yes, it's a good idea, we can send implmentationDesired + _handlerDesired to the contract and use _handlerDesired in the init call to proxy, since we don't have a function to update handler.
> 
> or we can introduce method to update handler and in this case we will be able to deploy proxy, init it with the old data and then update implem and handler from backend but I think just updating from the contract involves less friction

let's take a note of this now and keeping 1st option as way to go. 
there is an authorized method to update the handler also ( function setFallbackHandler(address handler) public authorized { ) 
they will be functional,  just old and can be changed. I will emit implementation, initialiser and index details so backend knows this always for detecting / deploying new smart account with same address on other chain.

tomarsachin2271:
> > _implementationUsedForDeployment
> > -implmentationDesired
> > _handlerUsedForDeployment
> > _handlerDesired
> > Also keeping in mind that we can't leave proxy uninitialized
> 
> yes, it's a good idea, we can send implmentationDesired + _handlerDesired to the contract and use _handlerDesired in the init call to proxy, since we don't have a function to update handler.
> 
> or we can introduce method to update handler and in this case we will be able to deploy proxy, init it with the old data and then update implem and handler from backend but I think just updating from the contract involves less friction

If we pass _implmentationDesired and _handlerDesired as argument to function and they are not involved in initial proxy deployment, then its easy to front run this transaction and someone can set another desired implementation & handler....

So if we do it this way, any parameter that's not involved in Proxy deployment, has to be part of user signature which needs to be verified in SCW.

livingrockrises:
> > > I now think it would be the same thing without any benefit. we either pass handler and owner or pass initialiser data and in both case they have to be part of the salt. in case of initializer we'd have to remember that to create a salt, in current way we have to remember handler to regenerate same proxy address.
> > 
> > 
> > Yep, I agree here. What I'am trying to say, that passing initializer data is a more universal approach. If we use initializer, even if `init` args or even name will change in the new implementation, we won't need to deploy new factory.
> > owner and handler will be the part of initializer (encoded args of the init function) and the salt will depend on initializer and thus on owner and handler
> 
> I see, good idea. I will make this change and update the test cases.

Done. https://github.com/bcnmy/scw-contracts/commit/82730da1407bf4f93ff92edd8cb5f8e5f447452a

filmakarov:
> If we pass _implmentationDesired and _handlerDesired as argument to function and they are not involved in initial proxy deployment, then its easy to front run this transaction and someone can set another desired implementation & handler....
> So if we do it this way, any parameter that's not involved in Proxy deployment, has to be part of user signature which needs to be verified in SCW.

that's right, thank you! 

And this gave me an idea.
What if we create an intermediary implementation with 2 functions:
1) `checkSignature` - regular signature verification, that signature has been issued by an address that is provided as owner
_implmentationDesired and _handlerDesired are part of the signature
2) `init`  - the function that after signature verification sets handler and implementation to _handlerDesired and _implmentationDesired

thus we introduce standardized flow :

1) we always deploy with a temporary placeholder impementation , that allows us to get same addresses for user
2) we call init of the temp implementation
3) the we call init of the full-scale implementation to setup the rest 

@livingrockrises @tomarsachin2271 thoughts?

filmakarov:
And I want to clarify this as well

```
if eq(calldataload(0), 0xee9118f700000000000000000000000000000000000000000000000000000000) {
                mstore(0, target)
                return(0, 0x20)
            }
```

this checking calldata costs us 32 gas for **every call** to the proxy

Why can't we just have a method getImplementationAddress() external view in implementation? that just returns result of sload(address()

livingrockrises:
> And I want to clarify this as well
> 
> ```
> if eq(calldataload(0), 0xee9118f700000000000000000000000000000000000000000000000000000000) {
>                 mstore(0, target)
>                 return(0, 0x20)
>             }
> ```
> 
> this checking calldata costs us 32 gas for **every call** to the proxy
> 
> Why can't we just have a method getImplementationAddress() external view in implementation? that just returns result of sload(address()

I can add this getImplementation() read method in the proxy and remove IProxy interface.

livingrockrises:
> > If we pass _implmentationDesired and _handlerDesired as argument to function and they are not involved in initial proxy deployment, then its easy to front run this transaction and someone can set another desired implementation & handler....
> > So if we do it this way, any parameter that's not involved in Proxy deployment, has to be part of user signature which needs to be verified in SCW.
> 
> that's right, thank you!
> 
> And this gave me an idea. What if we create an intermediary implementation with 2 functions:
> 
> 1. `checkSignature` - regular signature verification, that signature has been issued by an address that is provided as owner
>    _implmentationDesired and _handlerDesired are part of the signature
> 2. `init`  - the function that after signature verification sets handler and implementation to _handlerDesired and _implmentationDesired
> 
> thus we introduce standardized flow :
> 
> 1. we always deploy with a temporary placeholder impementation , that allows us to get same addresses for user
> 2. we call init of the temp implementation
> 3. the we call init of the full-scale implementation to setup the rest
> 
> @livingrockrises @tomarsachin2271 thoughts?

Hmm I had thought of keeping v0.0.0 implementation immutable in proxy. That was it's pointing to something reliable since birth of it and owner storage slot is set appropriate. 

you mean to have this ^ implementation = intermediatory implementation to have above methods? checkSignature and init (rather sets handler and implementation) 

but none of these methods or params like _handlerDesired, _implementationDesired would be in deploy method of wallet factory right?
it's basically still two step as our approach 1. but something to think about..

livingrockrises:
> > And I want to clarify this as well
> > ```
> > if eq(calldataload(0), 0xee9118f700000000000000000000000000000000000000000000000000000000) {
> >                 mstore(0, target)
> >                 return(0, 0x20)
> >             }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this checking calldata costs us 32 gas for **every call** to the proxy
> > Why can't we just have a method getImplementationAddress() external view in implementation? that just returns result of sload(address()
> 
> I can add this getImplementation() read method in the proxy and remove IProxy interface.

done

AmanRaj1608 (approved):

livingrockrises:
merging this

filmakarov:
> I can add this getImplementation() read method in the proxy and remove IProxy interface.

I would suggest adding it not to the proxy, but to the implementation. Thi will make proxy deployment cheaper.

livingrockrises:
> > > I can add this getImplementation() read method in the proxy and remove IProxy interface.
> > 
> > 
> > I would suggest adding it not to the proxy, but to the implementation. Thi will make proxy deployment cheaper.
> 
> got it. I will change it in c4-remediations branch now

done. 
https://github.com/bcnmy/scw-contracts/commit/293efe2f919c636f3baad6c2b1ffce46780aa391

filmakarov:
> Hmm I had thought of keeping v0.0.0 implementation immutable in proxy. That way it's pointing to something reliable since birth of it and owner storage slot is set appropriate.
> 
> you mean to have this ^ implementation = intermediatory implementation to have above methods? checkSignature and init (rather sets handler and implementation)
> 
> but none of these methods or params like _handlerDesired, _implementationDesired would be in deploy method of wallet factory right? it's basically still two step as our approach 1. but something to think about..

Yes. that's exactly what I mean. Having minimal and reliable v0.0.0 implementation for all of the proxies at deploy be default, so it allows us to always have same address for newly deployed proxies.

step 1: we deploy proxy pointing to the v0.0.0 impl
step 2: we call preInit on v0.0.0. it verifies that _handlerDesired and _implementationDesired been signed by owner. if it has been signed properly, we can change implementation to _implementationDesired 
step 3: we call init of _implementationDesired to set owner and  _handlerDesired and other init stuff.
(btw in this case we have no need of recording handler and keeping it as a part of the salt. only owner is needed. we will just pass  _handlerDesired , that will be signed and verified in v0.0.0)

All the 3 steps should be completed in one txn.

I think this really can work and will allow us to solve this problem with keeping same address for user across chains

livingrockrises:
> > Hmm I had thought of keeping v0.0.0 implementation immutable in proxy. That way it's pointing to something reliable since birth of it and owner storage slot is set appropriate.
> > you mean to have this ^ implementation = intermediatory implementation to have above methods? checkSignature and init (rather sets handler and implementation)
> > but none of these methods or params like _handlerDesired, _implementationDesired would be in deploy method of wallet factory right? it's basically still two step as our approach 1. but something to think about..
> 
> Yes. that's exactly what I mean. Having minimal and reliable v0.0.0 implementation for all of the proxies at deploy be default, so it allows us to always have same address for newly deployed proxies.
> 
> step 1: we deploy proxy pointing to the v0.0.0 impl step 2: we call preInit on v0.0.0. it verifies that _handlerDesired and _implementationDesired been signed by owner. if it has been signed properly, we can change implementation to _implementationDesired step 3: we call init of _implementationDesired to set owner and _handlerDesired and other init stuff. (btw in this case we have no need of recording handler and keeping it as a part of the salt. only owner is needed. we will just pass _handlerDesired , that will be signed and verified in v0.0.0)
> 
> All the 3 steps should be completed in one txn.
> 
> I think this really can work and will allow us to solve this problem with keeping same address for user across chains

hmm I wanna see how code shapes up along with test cases, and also think about signature verification flow and not leaving proxy uninitialized for long, also keeping gas optimisations in mind. we can create new branch out of c4-remediations to test this theory. if done well it can really help client sdk and backend efforts. can keep a call for this sometime next week 
cc @talhamalik883 @tomarsachin2271

PR #39, opened 2023-02-20
fix mixedAuth modifiers privileges

AmanRaj1608:
- [x]  keep mixedAuth for setOwner() and write test cases to update it from owner and from sample social recovery module.
- [x]  setOwner() could also be done using AA flow…
- [x]  this social recovery module should not be able to call anything else than setOwner()
- [x]  instead of mixedAuth in setImpl, use _requireFromEntryPointOrOwner()
- [x]  write test cases for the above for userOp calldata to be updateImpl
- [x]  should not be able to set implementation from any module call, or execute() / execFromEntryPoint() method of AA flow
- [ ]  updateEntryPoint would be completely removed.
- [ ]  test case and use journey methods in the SDK to upgrade entry point (by updating implementation - as of now!)


contracts/smart-contract-wallet/test/SampleSocialRecoveryModule.sol
```diff
@@ -0,0 +1,110 @@
...
+                moduleOwner,
+                0,
+                data,
+                Enum.Operation.Call
+            ),
+            "Could not execute recovery"
+        );
+    }
+
+    function authCall(SmartAccount _account, bytes memory _data) public {
```
livingrockrises, line 79: shouldn't you remove this method or use it internally if needed? it opens access to everyone once this module is authorised in smart account

contracts/smart-contract-wallet/test/SampleSocialRecoveryModule.sol
```diff
@@ -0,0 +1,110 @@
...
+        friends = _friends;
+        threshold = _threshold;
+    }
+
+    function confirmTransaction(bytes32 dataHash) public onlyFriend {
+        require(!isExecuted[dataHash], "Recovery already executed");
+        isConfirmed[dataHash][msg.sender] = true;
+    }
+
+    function recoverAccess(SmartAccount owner, address newOwner)
```
livingrockrises, line 53: this should be called 'account' and not 'owner'
first argument...

contracts/smart-contract-wallet/test/SampleSocialRecoveryModule.sol
```diff
@@ -0,0 +1,110 @@
...
+    mapping(bytes32 => bool) public isExecuted;
+    // isConfirmed mapping maps data hash to friend's address to confirmation status.
+    mapping(bytes32 => mapping(address => bool)) public isConfirmed;
+
+    modifier onlyFriend() {
+        require(isFriend[msg.sender], "Method can only be called by a friend");
+        _;
+    }
+
+    function setup(
```
livingrockrises, line 25: this probably should not be a public method. anyone can be moduleOwner... what is the role of moduleOwner?

test/smart-wallet/testGroup1.ts
```diff
@@ -773,4 +775,131 @@ describe("Base Wallet Functionality", function () {
...
+    );
+    socialRecoveryModule = await SocialRecoveryModule.connect(
+      accounts[0]
+    ).deploy();
+    console.log("SocialModule deployed at ", socialRecoveryModule.address);
+
+    // setup social recovery module, set bob, charlie as friends and set threshold as 2
+    await socialRecoveryModule
+      .connect(accounts[0])
+      .setup([bob, charlie], 2, userSCW.address);
```
livingrockrises, line 798: this should probably be only smart account transaction. check candide contract for social recovery and transaction flow

contracts/smart-contract-wallet/test/SampleSocialRecoveryModule.sol
AmanRaj1608: removed

test/smart-wallet/testGroup1.ts
AmanRaj1608: updated!

contracts/smart-contract-wallet/test/SocialRecoveryModule.sol
```diff
@@ -42,83 +38,82 @@ contract SocialRecoveryModule {
...
-            newOwner
+    function recoverAccess(address payable _wallet, address _newOwner) public {
+        require(_onlyFriends(_wallet, msg.sender), "sender not a friend");
+        bytes32 recoveryHash = getRecoveryHash(
+            _wallet,
+            _newOwner,
+            walletsNonces[_wallet]
         );
-        bytes32 dataHash = getDataHash(data);
-        require(!isExecuted[dataHash], "Recovery already executed");
```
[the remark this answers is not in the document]
AmanRaj1608: Also as nonce is introduced so, dataHash can't be used again as it will after calling `recoverAccess`

contracts/smart-contract-wallet/test/SocialRecoveryModule.sol
```diff
@@ -0,0 +1,119 @@
...
+    mapping(address => mapping(address => bool)) public isFriend;
+
+    // isConfirmed - map of [recoveryHash][friend] to bool
+    mapping(bytes32 => mapping(address => bool)) public isConfirmed;
+    mapping(address => uint256) internal walletsNonces;
+
+    /**
+     * @dev Setup function sets initial storage of contract. Only by SCW owner.
+     */
+    function setup(address[] memory _friends, uint256 _threshold) public {
```
livingrockrises, line 23: later we can update this with wallet address as argument and onlyModule(_wallet) modifier such that setUp is called after module is enabled. 
currently module can be enabled later which is ok as per this logic

contracts/smart-contract-wallet/test/SocialRecoveryModule.sol
```diff
@@ -42,83 +38,82 @@ contract SocialRecoveryModule {
...
-    /// @param data Data payload.
     /// @return Data hash.
-    function getDataHash(bytes memory data) public pure returns (bytes32) {
-        return keccak256(data);
+    function getRecoveryHash(
+        address _wallet,
+        address _newOwner,
+        uint256 _nonce
+    ) public pure returns (bytes32) {
+        return keccak256(abi.encode(_wallet, _newOwner, _nonce));
```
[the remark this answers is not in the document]
livingrockrises: where is getRecoveryHash used?

I believe it only applies once you add domain separator logic to allow gasless calls from recoverers, in which case recoverAccess will also accept signature. perhaps confirmRecovery can be gasless the exact same way.
correct me if I'm wrong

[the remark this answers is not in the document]
livingrockrises: we can remove then getRecoveryHash for now if it's not part of test flow which it shouldn't be imo. add it later when gasless actions by guardians can added altogether..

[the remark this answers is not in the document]
AmanRaj1608: Currently using this `getRecoveryHash` as a process to verify friend/guardian.

Wallet calls setup, adds friends and thersold.
Friend calls confirmRecovery, and a hash is generated that for a wallet, newOwner, nonce. a confirmation has been made.
Other friends confirms the same.
Then any one friend can call the revovery call in which we check if that hash has enough confirmations and confirm the call.

contracts/smart-contract-wallet/test/SocialRecoveryModule.sol
AmanRaj1608: setup is called by SCW only as the friends is stored with the `msg.sender`, but got your point

contracts/smart-contract-wallet/test/SocialRecoveryModule.sol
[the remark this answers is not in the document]
livingrockrises: I see how it's used.

livingrockrises (approved):
looks good with current version of social login module logic

PR #40, opened 2023-02-21
bump solidity 0.8.17 version

AmanRaj1608:
- It is recommend to use the latest version i.e. 0.8.17 instead of using 0.8.12
- Avoid floating pragmas for non-library contracts. While floating pragmas make sense for libraries to allow them to be included with multiple different versions of applications, it may be a security risk for application implementations. & It is recommended to pin to a concrete compiler version.
- Consider using the latest OZ packages (v4.8.1) within package.json. Many gas optimisation and fixes.


livingrockrises (approved):
LGTM

PR #42, opened 2023-02-22
C4 400 - EIP1271 compliance

filmakarov:
- Changed interface to (bytes32, bytes) instead of (bytes, bytes)
- Changed magic value to the new one
- Added preapproved hashes flow for the isValidSignature verification
- Test cases including 0x vulnerability test cases

As per OpenSea etc compatibility.
Per Chirag's advice I checked [Ambire Wallet ](https://wallet.ambire.com/) flow. It is possible to connect to OpenSea with Ambire wallet through wallet connect and using standard signed messages (not a preapproved one).

In OpenSea you sign with eoa owner of scw, but in the message there is an scw address
Screen Shot 2023-02-22 at 13 55 01

After researching [their code](https://github.com/AmbireTech/wallet/blob/f6c952c8b7ad0c0a5d48d7ba22ab265d8a5429d0/contracts/Identity.sol#L152-L159), it is clear that they use standard procedure with recovering signer and comparing it to eoa owner.
So it means our implementation will also be compatible with OpenSea's login flow.

As for making an order flow (signing typed 1271 data) , our wallet will also be compatible as : 
- OpenSea [use new magic value](https://github.com/wyvernprotocol/wyvern-v3/blob/4790c04604b8dc1bd5eb82e697d1cdc8c53d57a9/contracts/exchange/ExchangeCore.sol#L24) 
- they use standard isValidSignature  interface [ (1)](https://github.com/wyvernprotocol/wyvern-v3/blob/4790c04604b8dc1bd5eb82e697d1cdc8c53d57a9/contracts/exchange/ExchangeCore.sol#L185-L191)  [(2 - Lines 731-803)](https://etherscan.io/address/0x7f268357a8c2552623316e2562d90e642bb538e5#code)


test/smart-wallet/signatures/EIP1271Test.ts
```diff
@@ -0,0 +1,589 @@
...
+
+    let dataHash = await smartAccountWithHandlerInterface.callStatic["getMessageHash(bytes)"](dataToSign);
+    
+    let eip1271MagicValue = "0x1626ba7e"; 
+    let value = await smartAccountWithHandlerInterface.callStatic["isValidSignature(bytes32,bytes)"](dataHash, "0x");
+    expect(value).to.be.equal(eip1271MagicValue);
+  
+  });
+
+  it("Fallback handler returns magic value if correct signature has been provided directly to Smart Account", async function () {
```
livingrockrises, line 313: didn't get this case and how it works?
this is exactly same as ambire x opensea ux we checked right?

test/smart-wallet/signatures/EIP1271Test.ts
```diff
@@ -0,0 +1,589 @@
...
+
+  await expect (signerSmartAccount.connect(accounts[1]).execTransaction(
+    transaction,
+    refundInfo,
+    fakeSignature
+  )).to.be.reverted;
+  
+});
+
+it("0x exploit 1271 | Reverts if isValidSignature changes the state", async function () {
```
livingrockrises, line 490: just curious what was the exploit path and fix for this case?

test/smart-wallet/signatures/EIP1271Test.ts
```diff
@@ -0,0 +1,589 @@
...
+        manipulatedSignature
+      )
+    ).to.be.revertedWith("BSA024");
+
+    expect(await token.balanceOf(charlie)).to.equal(0);
+
+  });
+
+  
+it("0x exploit 1271 | Reverts if trying to use 1271 signature instead of EOA signature", async function () {
```
livingrockrises, line 438: I guess we don't need to make change like this commit? https://github.com/0xProject/0x-monorepo/commit/2171632e1e0f83ca55d7cc0072e3fb6d658219a0

livingrockrises:
Great!, all cases looks good to me. Just left some comments for knowledge / discussions. 

later steps ahead would be to write wallet connect client for our smart account and try to test UX on opensea. 

For the test cases we can write something with wyvern contracts that has opensea list-for-sale flow

contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -1,31 +1,31 @@
...
+import {SmartAccountErrors} from "./common/Errors.sol";
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
 
 contract SmartAccount is 
      Singleton,
      BaseSmartAccount,
-     IERC165,
      ModuleManager,
+     FallbackManager,
```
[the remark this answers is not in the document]
filmakarov: No problem, we just should always maintain the correct order. And it is verified by autotests. If order changes and storage layout changes, tests will fail and we can fix the layout in the lib

contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -213,7 +219,9 @@ contract SmartAccount is
                 );
             // Execute transaction.
             txHash = keccak256(txHashData);
-            checkSignatures(txHash, txHashData, signatures);
+
+            //checkSignatures(txHash, txHashData, signatures);
```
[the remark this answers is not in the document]
filmakarov: great! removing.

contracts/smart-contract-wallet/handler/DefaultCallbackHandler.sol
```diff
@@ -5,13 +5,47 @@ import "../interfaces/ERC1155TokenReceiver.sol";
...
 
 /// @title Default Callback Handler - returns true for known token callbacks
 /// @author Richard Meissner - <richard@gnosis.pm>
-contract DefaultCallbackHandler is ERC1155TokenReceiver, ERC777TokensRecipient, ERC721TokenReceiver, IERC165 {
+contract DefaultCallbackHandler is ERC1155TokenReceiver, ERC777TokensRecipient, ERC721TokenReceiver, IERC165, ISignatureValidator {
     string public constant NAME = "Default Callback Handler";
     string public constant VERSION = "1.0.0";
 
+    //keccak256(
+    //    "SmartAccount(bytes message)"
```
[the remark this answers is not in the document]
filmakarov: +

contracts/smart-contract-wallet/handler/DefaultCallbackHandler.sol
```diff
@@ -5,13 +5,47 @@ import "../interfaces/ERC1155TokenReceiver.sol";
...
+        // Caller should be a SmartAccount
+        SmartAccount smartAccount = SmartAccount(payable(msg.sender));
+
+        if (_signature.length == 0) {
+            return (smartAccount.signedMessages(_dataHash) != 0) ? EIP1271_MAGIC_VALUE : bytes4(0);
+        } else {
+            try smartAccount.checkSignatures(_dataHash, _signature) {
+                return EIP1271_MAGIC_VALUE;
+            } catch {
+                return bytes4(0);
```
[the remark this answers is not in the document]
filmakarov: I saw both in different implementations. I can change to 0x0xffffffff

test/smart-wallet/signatures/EIP1271Test.ts
filmakarov: yep. it's the case, when some off-chain script calls isValidSignature, not other contract

test/smart-wallet/signatures/EIP1271Test.ts
filmakarov: here's the description https://samczsun.com/the-0x-vulnerability-explained/

I don't feel it was applicable to us since 0x implementation relied on true/false return of signature validation process, not a magic value based.

however, it was interesting to implement this

test/smart-wallet/signatures/EIP1271Test.ts
filmakarov: I think , we don't. as tests show that isValidSignature calls to eoa revert

filmakarov:
> later steps ahead would be to write wallet connect client for our smart account and try to test UX on opensea.
> For the test cases we can write something with wyvern contracts that has opensea list-for-sale flow

I agree, it would be interesting to see how it will actually work with opensea.

test/smart-wallet/signatures/EIP1271Test.ts
livingrockrises: interesting

test/smart-wallet/signatures/EIP1271Test.ts
livingrockrises: there isn't much mentioned on validator changing the state

livingrockrises (approved):
approved

PR #54, opened 2023-03-09
Wallet factory redesign

livingrockrises:
As per the discussion, this change can ease the deployment of SCW on multiple chains while keeping address the same. In other words without needing to track the information like implementation, initialiser(handler address etc) to deploy on another chain.

For any case we would be upgrading to new implementation
i. enforced : batch when user sends first userop, when user sends userop and upgrade is necessary (implementation change due to entry point change etc)
ii. informed: by dapps and giving them ability to just call setImplementation etc

Benefits:

All the wallets will be deployed at the same address for same owner and index(0,1,2..)
Deployed proxy will have accurate owner storage, minimal handler and functional audited secure implementation.

once deployed handler can be set new and implementation can be set new.

below is the case in which wallet deployment and upgrades can be done as part of single userop.

initcode : wallet factory + deployment data  
calldata: executeCall, executeBatchCall [ where batch is: self:updateImplementationAndCall(,), self:setFallbackHandler, self:updateImplementation, dapp:dappmethod ]

WIP.....

updating test cases as updateImplementation() is not requiredFromEntryPointOrOwner anymore, but it is mixedAuth. which means userOp.calldata should not be updateImplementation anymore but executeCall and _to:wallet, _data:updateImplementation function signature.

new entry point comes and implementation must be upgraded. above first deploy and call should go through prior EP and then onwards use new entrypoint!

methods in the sdk for setImplementation and setFallbackHandler etc and prepating userops..


contracts/smart-contract-wallet/SmartAccountFactory.sol
```diff
@@ -101,6 +115,8 @@ contract SmartAccountFactory {
         }
         require(address(proxy) != address(0), "Create call failed");
 
+        bytes memory initializer = getInitializer(_owner);
```
tomarsachin2271, line 118: Also emit a different event in this function as well.

contracts/smart-contract-wallet/utils/GasEstimatorSmartAccount.sol
```diff
@@ -9,15 +9,14 @@ contract GasEstimatorSmartAccount {
...
-        uint _index,
+        address _owner,
+        uint256 _index,
         bytes calldata _data // execTransaction data // counterFactual wallet should have assets if required
     ) external returns (bool success, bytes memory result, uint256 gas) {
         // solhint-disable
         uint256 initialGas = gasleft();
         address _wallet = SmartAccountFactory(_factory)
-            .deployCounterFactualWallet(_implementation, initializer, _index);
+            .deployCounterFactualWallet(_owner, _index);
```
tomarsachin2271, line 19: Do we call this method using eth_call?

tomarsachin2271 (commented):
Changes looks good. Added 2 comments.

contracts/smart-contract-wallet/utils/GasEstimatorSmartAccount.sol
livingrockrises: yes, this contract is for gas estimation test cases when wallet is not already deployed..

contracts/smart-contract-wallet/SmartAccountFactory.sol
livingrockrises: what would that event be?

tomarsachin2271: In deployCounterFactualWallet, emit event `CounterFactualAccountCreated`
and in deployWallet emit event `AccountCreated`

livingrockrises: I see. ok

PR #71, opened 2023-03-17
Fix ImplementationUpdated event arguments

filmakarov:
Was emitting proxy address instead of prev impl address

### DIFF
--- a/test/proxy-upgrades/enable-2D-nonces.ts
+++ b/test/proxy-upgrades/enable-2D-nonces.ts
@@ -33,7 +33,7 @@ export async function deployEntryPoint(
   return EntryPoint__factory.connect(epf.address, provider.getSigner());
 }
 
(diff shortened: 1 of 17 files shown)


test/proxy-upgrades/enable-2D-nonces.ts
```diff
@@ -193,7 +193,7 @@ describe("Upgradeability", function () {
 
     await expect(
       userSCW.connect(accounts[0]).updateImplementation(baseImpl5.address)
-    ).to.emit(userSCW, "ImplementationUpdated");
+    ).to.emit(userSCW, "ImplementationUpdated").withArgs(baseImpl.address, baseImpl5.address);
```
livingrockrises, line 196: ahaha nice thats why withArgs is important, won't ignore it next time when checking events

PR #77, opened 2023-03-22
Remade 1271 implementation, removed signedMessages flow

filmakarov:
* **Remade 1271 implementation:** 
1) isValidSignature is now in the SmartAccount itself which is more logical
2) if owner has `code.length > 0`, isValidSignature now forwards the request to owner.isValidSignature. That can be useful if the owner is a wallet or if the signature verification has been transferred to one of the modules. In this case it is expected that this module (like multisig) will be the SmartAccount owner 
* **Removed signedMessages flow**, as it is unlikely anyone will use this flow, when it is needed to send the txn to approve message first, instead of just providing the signature along with the isValidSignature call. It allows to remove one mapping from the Smart Account storage layout. Removed SignMessageLib as it is not needed anymore

added/removed test cases accordingly


contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -33,7 +34,8 @@ contract SmartAccount is
     ISignatureValidatorConstants,
     IERC165,
     ReentrancyGuard,
-    SmartAccountErrors
+    SmartAccountErrors,
+    ISignatureValidator
```
livingrockrises, line 38: while adding this you can remove ISignatureValidatorConstants

filmakarov:
> reviewed

thanks, addressed the comments

livingrockrises (approved):

PR #78, opened 2023-03-25
Factory deployment script

talhamalik883:
The PR includes factory contract deployment script.

@livingrockrises make sure to check these numbers

const deploymentGasPrice = 30e9 
// this number ^^ is 100 gwei in infitinitism repo. its high to make sure factory deployment trx never fails but i kept is as 30 as gas price currently is quite low. you can change it while deploying contract


livingrockrises (approved):
approving. do make values dynamic per network

PR #79, opened 2023-03-27
Deployment scripts refining

filmakarov:
It's a draft PR, to start the conversation.

1) Made some name changes to make reading the scripts more straightforward
2) Added framework to set individual gas price for deploying the Deployer Contract or take em from Hardhat config
3) Removed the flow to check if there's a DeployerContractInstance and deploy it if there's none. Now we expect this to be deployed for sure with the `deployer-contract.deploy.ts`

### Questions

1) will the `deploymentGasLimit` for the Deployer Contract (287000 gas  units) vary from chain to chain? Maybe we do not need to setup it individually for every chain, we can just set higher number, the rest is always refunded to the deployer contract
2) Why have we changed `CALLBACK_HANDLER` to `FallBACK_HANDLER`? And commented out callback handler deployment in the `deploy.ts` script?
3) I tested it for Mumbai and Fuji (avax testnet). Everything works smoothly, it deploys Deployer contract to the same address. By some reason it fails for Eth Goerli with 
```
  reason: 'cannot estimate gas; transaction may fail or may require manual gas limit',
  code: 'UNPREDICTABLE_GAS_LIMIT'
```
I think this is because of very high gas prices in Goerli rn


AmanRaj1608 (approved):

livingrockrises:
1. Yes gas limits can vary, i commented this in @talhamalik883 's PR earlier. I'm not sure about custom high limit, earlier this value was used to calculate amount of funds needs to be sent to deployer(deployer)  

2. resolved on dm (it's deployed from account factory now). we should fix the name fallback handler 

3. not sure, could be provider or gas price issue. we can catch and log

filmakarov:
> 1. Yes gas limits can vary, i commented this in @talhamalik883 's PR earlier. I'm not sure about custom high limit, earlier this value was used to calculate amount of funds needs to be sent to deployer(deployer)

I successfully deployed to all testnets yesterday, and 287000 worked fine for all of them.

> 3. not sure, could be provider or gas price issue. we can catch and log

I think it was goerli issue with gas, yesterday it worked smooth.

PR #80, opened 2023-03-27
support for future EP getNonce()

filmakarov:
According to [Proposal] (https://docs.google.com/document/d/1MywdH_TCkyEjD3QusLZ_kUZg4ZEI00qp97mBze9JI4k/edit#)
and @livingrockrises 's idea.

Realisation approved by Taek.

Test cases added as per Taek's recommendation.


contracts/smart-contract-wallet/BaseSmartAccount.sol
```diff
@@ -66,7 +66,10 @@ abstract contract BaseSmartAccount is IAccount, BaseSmartAccountErrors {
             revert CallerIsNotAnEntryPoint(msg.sender);
         validationData = _validateSignature(userOp, userOpHash);
         if (userOp.initCode.length == 0) {
-            _validateAndUpdateNonce(userOp);
+            (bool success, ) = address(entryPoint()).staticcall(
```
livingrockrises, line 69: if it's a success we just skip?

livingrockrises: alright because entrypoint will validate and update nonce on its own. 

Have we looked into sample wallet implementation changes / making suitable for gnosis as per test cases in the open PR https://github.com/eth-infinitism/account-abstraction/pull/247

contracts/smart-contract-wallet/SmartAccount.sol
```diff
@@ -206,19 +206,29 @@ contract SmartAccount is
...
+        if (batchId == 0) 
+            return nonce();
         return nonces[batchId];
     }
 
     /**
-     * @dev Standard interface for 1d nonces. Use it for Account Abstraction flow.
+     * @dev Returns a nonce from EntryPoint if EntryPoint supports it. Otherwise returns internal AA-flow nonce.
      */
     function nonce() public view virtual override returns (uint256) {
```
livingrockrises, line 223: at this point I am also thinking why limit keySpace 0
abi.encodeWithSignature("getNonce(address,uint192)", address(this), 0)
if it allows for bundlers to send parallel transactions

livingrockrises: that would require getNonce() method for AA and getForwardNonce() for nonERC4337 transactions (name inspired from other PR https://github.com/bcnmy/scw-contracts/pull/76 )

livingrockrises (commented):
added some comments

contracts/smart-contract-wallet/SmartAccount.sol
filmakarov: yes, EP's 2d nonces would require introducing separate method for forward flow.
It can be actually a good idea since we've discussed on yesterday's call that AA flow should be the default flow.

PR #82, opened 2023-03-31
Paymaster Redesign

filmakarov:
This is the PR to redesign Paymaster.

## **Main topics:**
1. Paymaster to support several Entry Points
2. Paymaster doesn't need nonces after Entry Points control nonces, not SCW

### Paymaster to support several Entry Points
**Problem statement:**
Now, EP is [immutable](https://github.com/eth-infinitism/account-abstraction/blob/bd8e11437d21c26968193b972ff511724a7d041a/contracts/core/BasePaymaster.sol#L19) in the BasePaymaster.
That means, when the new EP is deployed, we have to switch to a new Paymaster.
However, not all the smart contract wallet users will immediately upgrade to a new implementation with a new EntryPoint (if it is [immutable](https://github.com/eth-infinitism/account-abstraction/blob/bd8e11437d21c26968193b972ff511724a7d041a/contracts/samples/SimpleAccount.sol#L32) for the SCW as well) or just update EP if it is changeable
So DApps will have to support deposits in two paymasters at the same time that is I believe not a good developer experience.

**Solution:**
Allow paymaster to support several EPs at the same time and protect from replay attacks with hashing Entry Point address in the `getHash()`.

_Dror Tirosh:_
> re: paymaster to support multiple entrypoints.
> interesting idea, though the solution is not modify the entryPoint member (which is currently immutable, for efficiency) but instead replace _requireFromEntryPoint, but is virtual.
> This way, you can hash multiple entrypoints.

_@leekt :_
> Since your paymaster will need stake on entrypoint anyway it’s totally fine to use storage on paymaster
> ya i see no reason why you shouldn’t use mapping for this
> yup yup since getHash() has entrypoint’s address and chain id, it is totally fine

### Paymaster doesn't need nonces after Entry Points control nonces, not SCW
**Problem statement:**

 Why we introduced nonces: https://github.com/code-423n4/2023-01-biconomy-findings/issues/360 

Why paymaster nonces are not needed anymore:
_Dror Tirosh:_
> I don't think a nonce is really needed in paymayster.
> The reason for adding a nonce to the paymaster was to handle cases where the nonce was not unique: a wallet could support static nonce, and it could re-use the VerifyingPaymaster validation multiple times.
> But now nonces uniqueness is validated by entrypoint. So a paymaster no longer need to check for uniqueness by itself.

> NOTE: we 're currently in the process of updating the entryPoint, which will enforce uniqueness of nonce and userOpHash.
> This way, paymaster will no longer need to maintain its own sender/nonce/entrypoint mapping, since all those fields are hashed into the userOpHash


filmakarov:
I added a very quick prototype of how several EPs support can look like. 
Hashing is TODO yet.

livingrockrises:
If we keep active entry points, that will allow paymaster to accept validatePaymasterUserop() from one of the supported entry points. for that aspect this change is good so far. 

 I have concern on immutable entryPoint and another state is mapping of entry point that we are managing...
 
 can we discuss about what would be impact on this?
 
```
 function depositFor(address paymasterId) external payable nonReentrant {
        if (paymasterId == address(0)) revert PaymasterIdCannotBeZero();
        if (msg.value == 0) revert DepositCanNotBeZero();
        paymasterIdBalances[paymasterId] =
            paymasterIdBalances[paymasterId] +
            msg.value;
        entryPoint.depositTo{value: msg.value}(address(this));
        emit GasDeposited(paymasterId, msg.value);
    }
```

When a dapp deposits, our paymaster deposits to fixed immutable entrypoint. and when they want to withdraw it's only going to withdraw (our paymaster needs to be deposited there only) from fixed entry point

```
function withdrawTo(
        address payable withdrawAddress,
        uint256 amount
    ) public override nonReentrant {
        if (withdrawAddress == address(0)) revert CanNotWithdrawToZeroAddress();
        uint256 currentBalance = paymasterIdBalances[msg.sender];
        if (amount > currentBalance)
            revert InsufficientBalance(amount, currentBalance);
        paymasterIdBalances[msg.sender] =
            paymasterIdBalances[msg.sender] -
            amount;
        entryPoint.withdrawTo(withdrawAddress, amount);
        emit GasWithdrawn(msg.sender, withdrawAddress, amount);
    }
```

livingrockrises:
Now let's take this case by case.

if there is new entry point, we set it as one of the active entry points in paymaster 

1. Transaction request comes from a wallet that hasn't upgraded the entry point. (request could be to upgrade implementation itself or any transaction intended to go via gasless flow) 

entry point calls validateUserOp on wallet (wallet already trusts EP) then calls validatePaymasterUserOp (our singleton PM with EP mapping already trusts EP) and it's deposited on EP immutable so validatePaymasterPrepayment would work.

2. 
now let's say upgrade has happened (say above tx was Impl upgrade). our wallet has new implementation that has new EP. now in sdk to bundler we send transaction on this new EP.
new EP calls validateUserOp on wallet (wallet already trusts EP) [ works because paymasterAndData != '0x' (if it was wallet deposit flow wallets would need to deposit on new EP)  ], but thats not important, next thing new EP will call validatePaymasterUserOp (our singleton PM with EP mapping already trusts EP) but validatePaymasterPrepayment would fail because we haven't shifted our deposit from immutable EP to next one we support or next to next one. 

Trying to point gas accounting problem on EP and what would be impact on these functions deposit and withdraw.

livingrockrises:
from Fillip : 

We can just make a handy way for a dapp to manage balances between two paymasters along with the stats of what % of users have upgraded to new EP=New Paymaster and what % have not.

it's not a big deal if not upgraded wallets will have limited functionality.
However, how do we manage the upgrade itself, as it should go through paymaster as well. If we switch the EP in the Paymaster, they won't be able to upgrade via sponsored txn

if we just keep EP mutable, and change it, that will make the wallets that has not upgraded yet not capable to be used with the Paymaster

PR #84, opened 2023-04-10
update contracts rebased with 0.6.0

livingrockrises:
1. BaseSmartAccount implements nonce() method to get nonce from EntryPoint

```
/**
     * @return nonce the account nonce.
     * @dev This method returns the next sequential nonce.
     * @notice For a nonce of a specific key, use `entrypoint.getNonce(account, key)`
     */
    function nonce() public view virtual returns (uint256) {
        return entryPoint().getNonce(address(this), 0);
    }
```

2. Remove function _validateAndUpdateNonce from BaseSmartAccount and SmartAccount (and all it's versions used for tests)

3. Add below function in abstract class BaseSmartAccount.sol
   function _validateNonce(uint256 nonce) internal view virtual {
    } 
    
 4. Remove below snippet from validateUserOp
 
  ```
 if (userOp.initCode.length == 0) {	
            _validateAndUpdateNonce(userOp);	
        }
```

5. Remove senderNonce from VerifyingSingletonPaymaster

rationale: 

The reason for adding a nonce to the paymaster was to handle cases where the nonce was not unique: a wallet could support static nonce, and it could re-use the VerifyingPaymaster validation multiple times.
But now nonces uniqueness is validated by entrypoint. So a paymaster no longer need to check for uniqueness by itself

6. remove nonce() method from SmartAccount.sol (and all it's versions used for tests) because it's already implemented in abstract BaseAccount 

7. no need for senderPaymasterNonce while signing for VerifyingSingletonPaymaster

### DIFF
--- a/contracts/smart-contract-wallet/BaseSmartAccount.sol
+++ b/contracts/smart-contract-wallet/BaseSmartAccount.sol
@@ -42,9 +42,12 @@ abstract contract BaseSmartAccount is IAccount, BaseSmartAccountErrors {
 
     /**
      * @return nonce the account nonce.
-     * subclass should return a nonce value that is used both by _validateAndUpdateNonce, and by the external provider (to read the current nonce)
+     * @dev This method returns the next sequential nonce.
+     * @notice For a nonce of a specific key, use `entrypoint.getNonce(account, key)`
      */
-    function nonce() public view virtual returns (uint256);
+    function nonce() public view virtual returns (uint256) {
+        return entryPoint().getNonce(address(this), 0);
+    }
 
     /**
      * return the entryPoint used by this account.
@@ -65,9 +68,7 @@ abstract contract BaseSmartAccount is IAccount, BaseSmartAccountErrors {
         if (msg.sender != address(entryPoint()))
             revert CallerIsNotAnEntryPoint(msg.sender);
         validationData = _validateSignature(userOp, userOpHash);
-        if (userOp.initCode.length == 0) {
-            _validateAndUpdateNonce(userOp);
-        }
+        _validateNonce(userOp.nonce);
         _payPrefund(missingAccountFunds);
     }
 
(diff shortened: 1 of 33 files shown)


filmakarov:
All the testcases pass successfully after npm update!

filmakarov:
In the `SmartAccount.sol` , `execTransaction_S6W()` method
Should we change `nonces[1]++` => `nonces[0]++`

filmakarov:
Ok, I'll do it (it will involve changing some test cases as well)
And I will deploy the contracts

filmakarov:
Also, let's discuss our strategy for how do we swap `nonce` and `getNonce` later to comply with the standard interface that other wallets will implement.

1) If we are going to ship now with `getNonce(uint256 batchId)` , we will have to reimplement it in the next version using the argument as key.
But the key should actually be uin192 so we'll have to convert it from `uint256` to `uint192`. that should bring no problems btw.
For sequential nonces, we can additionally implement `getNonce()` with no arguments.

2) now we have just `nonce()` with no arguments
If we swap the purpose of `nonce` and `getNonce`, we'll have to add `nonce(uint256 batchId)` and override `nonce()` to whether revert or return `nonces[0]`, right?

So, looks like there should be no problems when we decide to upgrade.

livingrockrises:
> Also, let's discuss our strategy for how do we swap `nonce` and `getNonce` later to comply with the standard interface that other wallets will implement.
> 
> 1. If we are going to ship now with `getNonce(uint256 batchId)` , we will have to reimplement it in the next version using the argument as key.
>    But the key should actually be uin192 so we'll have to convert it from `uint256` to `uint192`. that should bring no problems btw.
>    For sequential nonces, we can additionally implement `getNonce()` with no arguments.
> 2. now we have just `nonce()` with no arguments
>    If we swap the purpose of `nonce` and `getNonce`, we'll have to add `nonce(uint256 batchId)` and override `nonce()` to whether revert or return `nonces[0]`, right?
> 
> So, looks like there should be no problems when we decide to upgrade.

hmm it's good to think about those. imo as long as sdks can manage by passing a specific "getNonce function name" as an argument for a wallet [ like they have done for gnosis safe ] even keeping nonce as it is should not be as problem. otherwise we can just override getNonce(uint256 batchId) to use EP.getNonce like you said. and original nonce() can be blocked and add getForwardNonce(). here we have to be mindful and use new BaseAccount class as well because currently nonce() is implemented in that abstract class.

tomarsachin2271 (approved):

PR #92, opened 2023-05-30
Stateless validation for deployment userOp

filmakarov:
### Issue details
As [discovered](https://github.com/eth-infinitism/bundler-spec-tests/pull/24#issuecomment-1565853588) by @leekt, it turned out not staked entities (including SAs) can not access [associated storage](https://eips.ethereum.org/EIPS/eip-4337#storage-associated-with-an-address) in course of validation stage of deployment user operation (`userOp.initCode.length !=0`).

It means, that although the deployment of an ownerless smart account + module setup (that access the associated storage) is possible when done through a staked factory, the validation of such a user operation will be reverted by bundlers if validateUserOp accesses the associated storage: [link1](https://github.com/eth-infinitism/bundler/blob/a1cab449a45c991215648fe77b10148cc6466d62/packages/bundler/src/parseScannerResult.ts#L276) and [link2](https://github.com/stackup-wallet/stackup-bundler/blob/5408be0e32a499801b4278718cc56bd51723af17/pkg/entrypoint/simulation/storageslots.go#L88). 

This is the case for our ownerless setup, as even the deployment user op validation happens through the Authorization module that is deployed and set up along with SA deployment.

### Mitigation
1) Add `validateDeploymentUserOp` method to the module. [Code](https://github.com/bcnmy/scw-contracts/blob/8c50ad652658acc2cf5f3fc9b3029f1899bb9687/contracts/smart-contract-wallet/modules/EOAOwnershipRegistryModule.sol#L49-L71)
This method mimics the regular validation algorithms implemented by the Auth module, except it does not access storage. Instead, it takes the public key of an authorized party (owner address in case of ECDSA Auth module) from the `userOp.initCode`. 
This information about the party that will be able to sign userOps will always be at the initCode as the Auth module should be set up in course of deployment userOp. This information can not be messed up by a malicious actor as this would lead to receiving another SA address.

2) Add a [temporary Implementation](https://github.com/bcnmy/scw-contracts/blob/2023-05-30-validateUserOp-SA-Deployment-issue/contracts/smart-contract-wallet/factory/SmartAccountTempImplementation.sol).
This implementation:
- [calls](https://github.com/bcnmy/scw-contracts/blob/8c50ad652658acc2cf5f3fc9b3029f1899bb9687/contracts/smart-contract-wallet/factory/SmartAccountTempImplementation.sol#L618) `module.validateDeploymentUserOp` instead of `module.validateUserOp`.
- [replaces itself with a permanent implementation](https://github.com/bcnmy/scw-contracts/blob/8c50ad652658acc2cf5f3fc9b3029f1899bb9687/contracts/smart-contract-wallet/factory/SmartAccountTempImplementation.sol#L625-L628) after validating the first (deployment) userOp.

Thus, the first userOp, that is deployment userOp, uses a validation algorithm that does not access associated storage.
All the next userOps use regular validation algorithm that accesses storage, but it is allowed to do so.

This approach adds ~2,600 gas to the deployment of a new SA via userOp.
An alternative approach could be to just include an additional `if (userOp.initCode.length !=0)` to the SmartAccount.validateUserOp in order to call `module.validateDeploymentUserOp` instead of `module.validateUserOp`. This approach would add 217 gas to **every** userOp validation. So after 12 userOps, this approach becomes more expensive than switching implementations. So the latter was favoured.

### Example of SA deployment userOp via new flow
[Link](https://github.com/bcnmy/scw-contracts/blob/8c50ad652658acc2cf5f3fc9b3029f1899bb9687/test/gas-benchmarks/Basic-Operations.specs.ts#L205-L312)

### TODO:
- Add SA.Deployment.specs.ts ⇒ deploy thru userOp; deploy directly and temporary implementation still can validate the first userOp even with initCode.length=0
- Rebuild `setupTests` fixture to use the new deployment process by default.


filmakarov:
Closing this as per discussion:
Screen Shot 2023-08-17 at 12 22 33

Since all the deployment tests via Bundler now pass even with SA accessing associated storage, the exception was made.
Thus this PR becomes efficiently outdated

PR #94, opened 2023-06-07
Forward Flow moved to Module

filmakarov:
The initial purpose was to get rid of `delegatecalls` for security reasons.
However, after considering discussions in the EIP-6900 group decided that `delegatecalls` should be access controlled instead of being removed.
`Delegatecalls` are currently possible for:
* Forward Flow txns (`execTransaction`)
* Modular flow (`execTransactionFromModule`)

Considering that:
* We will need access control for Modules anyways
* Forward Flow will be rarely to never used soon

decided to move Forward Flow to a separate module. 
This will allow for:
* Lighter and prettier codebase for `SmartAccount.sol`
* Unified access control over `delegatecalls` in the future through Modules access controls, as FF now goes through `execTransactionFromModule` as well.

Now the only way of making delegatecalls is through `execTransactionFromModule`. It's up to the user now to enable or not a potentially dangerous module that uses `Enum.Operation.delegatecall`.
In any case, such a module can be quickly disabled.
In the future can introduce access control for `delegatecalls` in `execTransactionFromModule` through Guards-like architecture.

Check [this file](https://github.com/bcnmy/scw-contracts/blob/Ownerless-SA-Modular-Forward-Flow/test/module/ForwardFlow.Module.ts) for more details on how FF txns are working now.

p.s. `SmartAccount` became now completely stateless, as both `owner` var and `nonces` mapping has been moved to the corresponding Modules and deprecated.


AmanRaj1608 (approved):
LGTM

contracts/smart-contract-wallet/base/ModuleManager.sol
```diff
@@ -150,17 +150,30 @@ abstract contract ModuleManager is
...
-        Enum.Operation operation
+        Enum.Operation operation,
+        uint256 txGas
     ) public virtual returns (bool success) {
         // Only whitelisted modules are allowed.
         if (msg.sender == SENTINEL_MODULES || modules[msg.sender] == address(0))
             revert ModuleNotEnabled(msg.sender);
         // Execute transaction without further confirmations.
-        success = execute(to, value, data, operation, gasleft());
-        if (success) {
```
livingrockrises, line 160: we won't be doing this success check and emitting events anymore?

contracts/smart-contract-wallet/test/SocialRecoveryModule.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.17;
 import "../SmartAccount.sol";
+import {Enum} from "../common/Enum.sol";
```
livingrockrises, line 4: where will it be used?

contracts/smart-contract-wallet/test/upgrades/Guards/DelegateCallTransactionGuard.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity 0.8.17;
 
 import {Enum} from "../../../common/Enum.sol";
 import {BaseGuard} from "./GuardManager.sol";
-import {Transaction, FeeRefund} from "../../../BaseSmartAccount.sol";
+import {Transaction, FeeRefund} from "../../../modules/ForwardFlowModule.sol";
```
livingrockrises, line 6: this guard is WIP?

contracts/smart-contract-wallet/test/upgrades/v1/BaseSmartAccountV1.sol
```diff
@@ -33,7 +33,7 @@ struct FeeRefund {
  * This contract provides the basic logic for implementing the IAccount interface: validateUserOp function
  * Specific account implementation should inherit it and provide the account-specific logic
  */
-abstract contract BaseSmartAccount is IAccount, BaseSmartAccountErrors {
+abstract contract BaseSmartAccount is IAccount, BaseSmartAccountErrorsV1 {
```
livingrockrises, line 36: version number is confusing... current smart account file version says 2.0.0 
can we also not keep Errors.sol as it is?

src/utils/execution.ts
```diff
@@ -409,27 +409,28 @@ export const buildSafeTransaction = (template: {
...
-export async function buildecdsaModuleAuthorizedForwardTx(
+export async function buildEcdsaModuleAuthorizedForwardTx(
   destinationContract: string,
   callData: string,
   smartAccount: Contract,
   smartAccountOwner: Signer & TypedDataSigner,
-  moduleAddress: string,
-  value: number | string = 0,
+  validationModuleAddress: string,
+  forwardFlowModule: Contract,
```
livingrockrises, line 418: I'm trying to understand how two modules work together?
like you init with ECDSA module 
and how does this another enabled module - Forward module work? do we have some basic flow diagram?
when will the module manager methods execTransactionFromModule() will be triggered?

test/Proxy.specs.ts
```diff
@@ -30,7 +30,7 @@ describe("NEW::: Proxy ", async () => {
       let proxy = await Proxy.deploy(implementationAddress);
       await proxy.deployed();
       proxy = await ethers.getContractAt("SmartAccount", proxy.address);
-      await expect(proxy.getChainId()).to.be.reverted;
+      await expect(proxy.getImplementation()).to.be.reverted;
```
livingrockrises, line 33: what is this test case about?

test/module/ForwardFlow.Module.ts
```diff
@@ -73,22 +89,23 @@ describe("NEW::: Smart Account Forward Flow", async () => {
...
       smartAccountOwner,
       ecdsaModule.address,
-      tokenAmountToTransfer.toString(),
+      forwardFlowModule,
+      tokenAmountToTransfer.toString()
     );
     
     await expect(
-      userSA.execTransaction_S6W(transaction, feeRefund, signature)
+      forwardFlowModule.execTransaction(userSA.address, transaction, feeRefund, signature)
```
livingrockrises, line 103: so since this is enabled module via module manager, it internally makes call smarTaccount.execTransactionFromModule

but module manager is also used to add validation plugins right, which plays a role in AA transaction. just trying to fully unerstand

livingrockrises (commented):
reviewed

contracts/smart-contract-wallet/base/ModuleManager.sol
filmakarov: yep. success is used as return value now. I don't see the need to emit `ExecutionFromModuleSuccess` or `ExecutionFromModuleFailure` since we have `ExecutionSucces/ExecutionFailure` in `Executor.execute`.
And this `ModuleTransaction` event consumes 10k+ gas, I was thinking how can we use it on backend and actually didn't find any reason that is worth 10k gas. 
In any case, we can easily add them back if at some point we need them

contracts/smart-contract-wallet/test/WhitelistModule.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.17;
 import "../SmartAccount.sol";
+import {Enum} from "../common/Enum.sol";
```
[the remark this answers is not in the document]
filmakarov: It was used in `authCall` so I had to keep this import. Idk if this PoC Whitelist module will be even used, maybe we will remove it after test cases are fully rebuilt/refactored.
In any case, this file is out of Audit scope.

contracts/smart-contract-wallet/test/upgrades/Guards/DelegateCallTransactionGuard.sol
filmakarov: Yes, it was PoC for Smart Account v1. But we had a call with you and Sachin and decided to leave it in tests by now.

contracts/smart-contract-wallet/test/upgrades/v1/BaseSmartAccountV1.sol
filmakarov: I had to create separate Errors file for the legacy V1 version of SA, as for the current SA version I removed several errors which are related to FF and are not used anymore in SmartAccount.sol

Version number [is 1.0.0 for SmartAccountV1.sol](https://github.com/bcnmy/scw-contracts/blob/7e4df55e10b82eb9fb9b9aba9774fde76b9d5706/contracts/smart-contract-wallet/test/upgrades/v1/SmartAccountV1.sol#L42)

src/utils/execution.ts
filmakarov: So, the transaction is initiated at `ForwardFlowModule` by calling `execTransaction`.
Like it was previously `SmartAccount.execTransaction_S6w`.

Then ForwardFlowModule [calls SA.isValidSignature](https://github.com/bcnmy/scw-contracts/blob/7e4df55e10b82eb9fb9b9aba9774fde76b9d5706/contracts/smart-contract-wallet/modules/ForwardFlowModule.sol#L95-L100) to validate the signature.
SA forwards this to the corresponding Validation Module, which is ECDSA module.
And returns validation success to ForwardFlowModule

Then ForwardFlowModule [calls SmartAccount.execTransactionFromModule](https://github.com/bcnmy/scw-contracts/blob/7e4df55e10b82eb9fb9b9aba9774fde76b9d5706/contracts/smart-contract-wallet/modules/ForwardFlowModule.sol#L120-L126) to actually execute the transaction.

[handlePayment](https://github.com/bcnmy/scw-contracts/blob/7e4df55e10b82eb9fb9b9aba9774fde76b9d5706/contracts/smart-contract-wallet/modules/ForwardFlowModule.sol#L163) has also been rebuilt to work through `execTransactionFromModule`

test/Proxy.specs.ts
filmakarov: it shows that if a proxy has been deployed with an EOA as implementation (not even through Factory), it won't be callable.

test/module/ForwardFlow.Module.ts
filmakarov: Yes, Module Manager can enables any module, be it the validation module (mostly used for AA flow, and just doing validation) or any other module, that can be called directly and can do execution through `execTransactionFromModule`

filmakarov:
> reviewed

Addressed the comments/questions, waiting for the feedback

contracts/smart-contract-wallet/test/WhitelistModule.sol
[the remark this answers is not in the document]
livingrockrises: cool

src/utils/execution.ts
livingrockrises: hmm so this is going to ECDSA auth module to verify it. 
https://github.com/bcnmy/scw-contracts/blob/7e4df55e10b82eb9fb9b9aba9774fde76b9d5706/contracts/smart-contract-wallet/modules/ForwardFlowModule.sol#L96

so essentially there is only one validation plugin / module. and here we required this auth module to call enableModule to enable Forward module.
I got confused because modules you enable (old gnosis way) can just call smart-account but they can have their own sig verification (which is forward validate signature here -> passed to main auth module :D ) [[ Also this sig here gets appended with forward module address as I saw , i think ]]

at the time there will only be one plugin? if yes/ even no how do you enable or replace auth modules?  [[ here you won't call enableModule right? like in init we said ok use this as auth module ]]

when we have social recovery module or session key module, they would be enabled via enableModule flow right? 
hence the other day I was saying modules are classified as auth/validation modules and execution modules (like we enabled forward) tell me If what i said makes sense?

livingrockrises (approved):
LGTM

src/utils/execution.ts
filmakarov: 1. There can be any amount of modules enabled via enableModule at the same time.
2. Any amount of enabled modules can be Authorization Modules.
3. In AA flow, the userOp.signature is appended with Authorization module address, that is intended to validate this signature. Thus there can be ECDSA Auth Module, Passkey Auth Module, Session key Auth module enabled at the same time for a SA and userOps can be signed with any of those modules. If they are signed with an EOA, the userOp.signature should be appended with an ECDSA Auth module address, so the validation flow will be forwarded to the ECDSA Auth Module. If userOp is signed with passkey, the Passkey Auth module address should be appended etc. Validation module address is extracted from sig [here](https://github.com/bcnmy/scw-contracts/blob/7e4df55e10b82eb9fb9b9aba9774fde76b9d5706/contracts/smart-contract-wallet/SmartAccount.sol#L264-L266).
4. So the Auth Modules do not need to call `execTransactionFromModule`. They just validate the userOp sig, and execution is handled by EntryPoint calling SA.executeCall or any other method. This is an efficient AA flow.
5. Old-fashioned Gnosis style modules (let's call them exec modules) are also able to interact with SA through `execTransactionFromModule`. They can have their own Txn validation logic, but they shouldn't be called in an userOp (coz if you're doing thru userOp you should just use AA flow with Auth Module). They should be called directly. So they can do their own validation logic (or forward it to SA like ForwardFlow module) and then call `execTransactionFromModule`. However, this flow is inefficient and is only needed for one use case, when user explicitly wants Forward Flow -> ForwardFlowModule.

PR #98, opened 2023-06-21
Session Manager Module + ERC20 Session Validation Module

filmakarov:
The `Session Manager Module` is an Authorization module. It implements `validateUserOp` function to receive validation calls from`Smart Account`.
It implements associated storage pattern via mapping `smartAccount.address => SessionStorage`.
`SessionStorage` stores a root of the Merkle Tree, where leafs are Session objects, that contain `validUntil`, `validAfter`, address of the Session Validation module, and Session Data.
Session Data contains the Session Key that is authorized to sign for this Session, and the Session Parameters, specific to the Session Validation Module.

Session Validation Modules, such as `ERC20SessionValidationModule`, contain logic to check the permissions for various Session Key use cases.
For example, for `ERC20SessionValidationModule` permissions are: 
* address of the token to be transferred
* recipient address
* maximum amount to be transferred per one `userOp`
Session Validation Module also checks was the `userOp` signed by the authorized Session Key.

To authorize a new Session Key, it is required to recalculate the Merkle Tree and send a userOp that sets the new Merkle Root into the `SessionStorage`.

Session Key Manager Module => [link](https://github.com/bcnmy/scw-contracts/blob/SCW-V2-Ownerless-SA-SessionKey-Module/contracts/smart-contract-wallet/modules/SessionKeyModule.sol)
ERC20 Transfers Session Validation Module => [link](https://github.com/bcnmy/scw-contracts/blob/SCW-V2-Ownerless-SA-SessionKey-Module/contracts/smart-contract-wallet/modules/SessionValidationModules/ERC20SessionValidationModule.sol)
Test specs: [link](https://github.com/bcnmy/scw-contracts/blob/SCW-V2-Ownerless-SA-SessionKey-Module/test/module/SessionKeyManager.Module.specs.ts), [link](https://github.com/bcnmy/scw-contracts/blob/SCW-V2-Ownerless-SA-SessionKey-Module/test/module/SessionValidationModules/ERC20SessionValidation.Module.specs.ts)


contracts/smart-contract-wallet/modules/SessionKeyModule.sol
```diff
@@ -0,0 +1,101 @@
...
+struct SessionStorage {
+    bytes32 merkleRoot;
+}
+
+contract SessionKeyManager is BaseAuthorizationModule {
+    /*
+     * @dev mapping of Smart Account to a SessionStorage
+     * Info about session keys is stored as root of the merkle tree built over the session keys
+     */
+    mapping(address => SessionStorage) internal userSessions;
```
livingrockrises, line 18: would we need to retrieve merkle root in the backend service? 
asking if need to make this public or there is a read method, if applicable

livingrockrises: ok, resolved.

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    function getSessionKeys(
+        address smartAccount
+    ) external view returns (SessionStorage memory) {
+        return userSessions[smartAccount];
+    }
+
+    function setMerkleRoot(bytes32 _merkleRoot) external {
+        _setSessionData(msg.sender, _merkleRoot);
```
livingrockrises, line 27: so we will make SmartAccount call this method? 
what would the transaction flow look like from EntryPoint?

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
```diff
@@ -0,0 +1,101 @@
...
+                validUntil,
+                validAfter,
+                sessionValidationModule,
+                sessionKeyData
+            )
+        );
+        if (
+            !MerkleProof.verify(merkleProof, sessionKeyStorage.merkleRoot, leaf)
+        ) {
+            revert("SessionNotApproved");
```
livingrockrises, line 68: can validateUserOp revert?

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
```diff
@@ -0,0 +1,101 @@
...
+        }
+        return
+            _packValidationData(
+                //_packValidationData expects true if sig validation has failed, false otherwise
+                !ISessionValidationModule(sessionValidationModule)
```
livingrockrises, line 73: do we need to handle reverts here by making low level call and then control flow?

contracts/smart-contract-wallet/modules/SessionValidationModules/ISessionValidationModule.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {UserOperation} from "../BaseAuthorizationModule.sol";
+
+interface ISessionValidationModule {
+    function validateSessionUserOp(
+        UserOperation calldata _op,
+        bytes32 _userOpHash,
+        bytes calldata _data,
```
livingrockrises, line 10: can we rename these vars to more readable values?
also if you can check missing natspecs for all new contracts and functions

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
livingrockrises: I see it may revert with AA23 reverted but should we just treat it as failed sig?

livingrockrises (commented):
reviewed

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
filmakarov: EP=>SA.executeCall=>SKManager.setMerkleRoot
it's illustrated here 
https://github.com/bcnmy/scw-contracts/blob/19d046210f70557ccb31f8564658778b63890e42/test/module/SessionKeyManager.Module.specs.ts#L98-L132

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
filmakarov: I was thinking about it and concluded that this error doesn't relate to the signature itself, so I decided to keep the revert here as well as keep reverts in ERC20SessionValidationModule.
https://github.com/bcnmy/scw-contracts/blob/19d046210f70557ccb31f8564658778b63890e42/contracts/smart-contract-wallet/modules/SessionValidationModules/ERC20SessionValidationModule.sol#L34-L56

However, just returning SIG_VALIDATION_FAILED can be an option as well

contracts/smart-contract-wallet/modules/SessionValidationModules/ISessionValidationModule.sol
filmakarov: yes! already todo-ed adding natspec. will rename as well

filmakarov:
should I rename this folder to modular-smart-account?
Screen Shot 2023-06-22 at 16 41 31

contracts/smart-contract-wallet/modules/SessionValidationModules/ISessionValidationModule.sol
filmakarov: done

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
livingrockrises: hmm we could return SIG_VALIDATION_FAILED from validation module or catch the error in validateUserOp and return 1. because currently it would just revert even with say wrong passkey/ecdsa sig and won't allow backend systems  as bundler node to call simulateHandleOp with dummysig etc

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
livingrockrises: hmm to think about it, it makes sense since it's not technically signature failure. not sure how bundlers treat AA23 and if this could be a blocker for estimating user operations going on new ownerless account

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
livingrockrises: got it, resolved.

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
filmakarov: The solution here would be to rename SIG_VALIDATION_FAILED to just VALIDATION_FAILED

filmakarov: @arcticfloyd1984 ^^^

filmakarov: >currently it would just revert even with say wrong passkey/ecdsa sig

nope. In case of a wrong ecdsa sig it won't revert. in this case it will return SIG_VALIDATION_FAILED

https://github.com/bcnmy/scw-contracts/blob/a6ed50f2fdb317a6fcf5eff2e85749838fe12ce2/contracts/smart-contract-wallet/modules/SessionValidationModules/ERC20SessionValidationModule.sol#L58-L63

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
filmakarov: If we return SIG_VALIDATION_FAILED from validateUserOp, EntryPoint will revert with `FailedOp(opIndex, "AA24 signature error")`
If we revert in validateUserOp, EntryPoint will revert with `FailedOp(opIndex, string.concat("AA23 reverted: ", revertReason))`

In both cases, this is the revert from EP with the same `error FailedOp(uint256 opIndex, string reason)`
Just with different arguments

so, for the bundlers it should be the same, it is just validation that fails

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
livingrockrises: > The solution here would be to rename SIG_VALIDATION_FAILED to just VALIDATION_FAILED Then it will be good to return it in any case when validation fails
> 
> however, it makes things less informative. as it provides no information about the failure reason. that's why I decided to revert in ERC20SessionValidationModule and also revert with "SessionNotApproved" in SessionKeyManager

hmm I think then in this case we can leave it with reverts which will be captured as AA23. I just realised in paymasters repo and flow also there is AA33 expected for all cases not related to sig verification.

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
filmakarov: they have this 
```
catch Error(string memory revertReason) {
            revert FailedOp(opIndex, string.concat("AA23 reverted: ", revertReason));
        } catch {
            revert FailedOp(opIndex, "AA23 reverted (or OOG)");
```
in EntryPoint.sol, so they expect  that `validateUserOp` may revert, not only return SIG_VALIDATION_FAILED

livingrockrises:
please resolve the conflicts @filmakarov

livingrockrises (approved):
approved

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
filmakarov: > won't allow backend systems as bundler node to call simulateHandleOp with dummysig etc

however, EP.handleOps will revert in both cases, see [here](https://github.com/bcnmy/scw-contracts/pull/98#discussion_r1239682277)

so again, it makes no difference in terms of simulating handleOps whether to return sig_validation_failed or revert, they both lead EP to revert

contracts/smart-contract-wallet/modules/SessionKeyModule.sol
livingrockrises: > If we return SIG_VALIDATION_FAILED from validateUserOp, EntryPoint will revert with `FailedOp(opIndex, "AA24 signature error")` If we revert in validateUserOp, EntryPoint will revert with `FailedOp(opIndex, string.concat("AA23 reverted: ", revertReason))`
> 
> In both cases, this is the revert from EP with the same `error FailedOp(uint256 opIndex, string reason)` Just with different arguments
> 
> so, for the bundlers it should be the same, it is just validation that fails

yes I also checked in entry point code. in case of paymaster it is similar it would anyway revert in case of SIG_VALIDATION_FAILED as well with AA34 signature error and if validatePaymasterUserOp reverts it would revert with AA33 + error. 

the idea is and this is what I was looking for - it should not revert in simulateHandleOp method in the case it is merely SIG_VALIDATION_FAILED because the purpose is, for bundlers to be able to still estimate gas limits with dummy signatures.

PR #102, opened 2023-07-28
Multi(chain) userOp Validator

filmakarov:
This branch features an ECDSA validator that can validate any userOp from a list of userOps represented by a root of the Merkle tree.

- Initially it was planned to make a multichain validator: only chain specific parameters such as chainId, nonce, gas etc could vary
- However, for flexibility and gas efficiency, I decided to use the whole userOpHash-es as leaves of the Merkle Tree.
- This effectively makes this module not just a _Multichain_ Validator, but a _**Multi UserOp**_ Validator, since now even userOp.calldata may vary. 
- Of course , if the calldata is the same, this module can be used for multichain deployment + userOp processing.
- So with this module user can only sign once for _any_ amount of userOps. It significantly improves UX for crosschain dapps, allowing for example deployment of SA and enabling the first Session Key on any number of chains with only one signature required.
- This module can work as single chain ECDSA validator as well.

Screen Shot 2023-07-28 at 16 02 28

Fixes SMA-1


contracts/smart-account/modules/MultichainValidator.sol
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {BaseAuthorizationModule, UserOperation, ISignatureValidator} from "./BaseAuthorizationModule.sol";
```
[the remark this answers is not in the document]
filmakarov: fixed, thank you

contracts/smart-account/modules/MultichainValidator.sol
```diff
@@ -0,0 +1,85 @@
...
+
+        if (moduleSignature.length == 65) {
+            //it's not a multichain signature
+            return
+                _verifySignature(
+                    userOpHash,
+                    moduleSignature,
+                    address(uint160(sender))
+                )
+                    ? VALIDATION_SUCCESS
```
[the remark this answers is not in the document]
filmakarov: added

filmakarov:
fixed both issues

contracts/smart-account/modules/MultichainValidator.sol
```diff
@@ -47,27 +39,36 @@ contract MultichainECDSAValidator is EcdsaOwnershipRegistryModule {
```
tomarsachin2271, line 39: Add the validUntil and validAfter field in non-multichain signatures as well.

contracts/smart-account/modules/MultichainValidator.sol
```diff
@@ -47,27 +39,36 @@ contract MultichainECDSAValidator is EcdsaOwnershipRegistryModule {
...
                 multichainSignature,
                 address(uint160(sender))
             )
-                ? VALIDATION_SUCCESS
+                ? _packValidationData(
+                    false, //sigVerificationFailed = false
+                    validUntil == 0 ? type(uint48).max : validUntil,
+                    validAfter
+                )
                 : SIG_VALIDATION_FAILED;
```
tomarsachin2271, line 72: why we are not returning packed validation data in case of failed signature validation? Is it not used in EntryPoint ?

filmakarov: in case signature validation failed, there's no need to spend gas and pack validUntil and validAfter. they do not matter if the sig itself is wrong.
thus, having SIG_VALIDATION_FAILED value just 1 and returning it is enough. 
https://github.com/eth-infinitism/account-abstraction/blob/12be13e2e97b763e1ef294602b3f2072bc301443/contracts/samples/SimpleAccount.sol#L104

contracts/smart-account/modules/MultichainValidator.sol
filmakarov: We currently do not use validUntil and validAfter for single chain (regular) ecdsa signatures.

I suggest not to do it. 

1) In most cases it is not needed, as the signature is over the userOp hash and the userOp is immediately sent to the bundler to be handled (like txns by EOA, which are signed and immediately sent to mempool), so in 99% of cases I suppose that validUntil and validAfter will be expected to be unlimited.

2) it is not expected that validUntil and validAfter are provided from outside now. As per 4337 single chain signatures are expected to be over userOpHash. If we need to provide validUntil and validAfter from outside we will have to append it to userOpHash and sign, so they can't be changed. Then we will need to unpack them on-chain and spend gas for it.

3) IMO, the idea for validUntil and validAfter is that it is not just passed to the SA from outside, but is defined by the SA (or in our case validation module) itself. Depending on the module specific.
For example , that's how it is done is Account Recovery Module
First we detect validity timeframe using some module specific logic => https://github.com/bcnmy/scw-contracts/blob/8fbe98723639dd0f21c6703de8149746f1d65ed2/contracts/smart-account/modules/SocialRecoveryModule.sol#L211-L219
Then we pack it => https://github.com/bcnmy/scw-contracts/blob/8fbe98723639dd0f21c6703de8149746f1d65ed2/contracts/smart-account/modules/SocialRecoveryModule.sol#L250-L253

tomarsachin2271: If user doesn't sign validUntil and validAfter, it opens up a situation where user signature is valid indefinitely, if some bundler decides to hold off an userOp (could be a swap userOp) and decides to execute it later (may be when it gives max returns to bundler somehow).. the user account could be affected in a wrong way.

livingrockrises:
as bundler integration is merged, you need to resolve some conflicts here

contracts/smart-account/modules/MultichainValidator.sol
filmakarov: The same could have been valid for pre-AA txns/mempool. Also such a censorship may occur not only because of the economical interest of the bundler/validator, but for other reasons as well. 
However, since there are many actors, this system is censorship-resistant. 
The same should be valid for AA userOps pool.

According to this writeup by Yoav, 
https://notes.ethereum.org/@yoav/unified-erc-4337-mempool
One of the conditions of making erc4337 infra censorship-resistant
is that "Users can propagate user-operations to a network where any bundler can serve them."

What I'm trying to say is that if there is no censorship-resistant system, featuring the network of independent bundlers, just having validUntil/validAfter won't be enough to protect users from userOps censorship.

Thus, I believe that such a network of independent Bundlers will be in place, and if user chooses to use the private bundlers, those bundlers will be selected with a criteria of not profiting on their users (like flashbots do not benefit from txns processed).

Assuming this I still suggest not to introduce validUntil/validAfter for regular ecdsa signatures over userOpHash'es, which act more like a standard private key signatures over pre-AA txn.

filmakarov:
Conflicts resolved

filmakarov:
added descriptions and natspecs

test/bundler-integration/module/MultichainValidator.Module.specs.ts
```diff
@@ -0,0 +1,303 @@
...
+    await deployments.fixture();
+
+    const entryPoint = await getEntryPoint();
+    const smartAccountFactory = await getSmartAccountFactory();
+    const mockToken = await getMockToken();
+    const ecdsaModule = await getEcdsaOwnershipRegistryModule();
+    
+    const MultichainECDSAValidator = await ethers.getContractFactory("MultichainECDSAValidator");
+    const multichainECDSAValidator = await MultichainECDSAValidator.deploy();
+    const sessionKeyManager = await (await ethers.getContractFactory("SessionKeyManager")).deploy();
```
livingrockrises, line 68: are these being used in this test case anywhere?

contracts/smart-account/modules/MultichainECDSAValidator.sol
```diff
@@ -0,0 +1,102 @@
...
+                _verifySignature(
+                    userOpHash,
+                    moduleSignature,
+                    address(uint160(sender))
+                )
+                    ? VALIDATION_SUCCESS
+                    : SIG_VALIDATION_FAILED;
+        }
+
+        //otherwise it is a multichain signature
```
livingrockrises, line 62: can we check the exact length of signature if we know?

livingrockrises (commented):
reviewed

contracts/smart-account/modules/MultichainValidator.sol
livingrockrises: if we're passing validUntil, validAfter we can use _packValidationData() from Helpers class (infinitism) regardless. (sig failed / sig passed)

tomarsachin2271: Got it.. yeah in a censorship-resistant system, we wouldn't need to worry about it.

contracts/smart-account/modules/SessionKeyManagerModule.sol
```diff
@@ -34,6 +46,7 @@ contract SessionKeyManager is BaseAuthorizationModule {
      */
     function setMerkleRoot(bytes32 _merkleRoot) external {
         userSessions[msg.sender].merkleRoot = _merkleRoot;
+        // TODO:should we add an event here? which emits the new merkle root
```
tomarsachin2271, line 49: Not sure how helpful it will be. give me an example where this merkle root alone will be helpful

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
```diff
@@ -3,6 +3,20 @@ pragma solidity 0.8.17;
 import "./ISessionValidationModule.sol";
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 
+/**
+ * @title ERC20 Session Validation Modul for Biconomy Smart Accounts.
```
tomarsachin2271, line 7: typo in Modul**e**

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
```diff
@@ -3,6 +3,20 @@ pragma solidity 0.8.17;
...
+ * @dev Validates ERC20 transfers (and approvals) session key signed userOps
```
tomarsachin2271, line 8: Can be better worded.

Validates userOps for ERC20 transfers and approvals using a session key signature.

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
```diff
@@ -3,6 +3,20 @@ pragma solidity 0.8.17;
...
+ * Performs basic verifications for every session key signed userOp.
+ * Checks if the session key has been enabled, that it is not due and has not yet expired
```
tomarsachin2271, line 10: These statements are true for SessionKeyManager, not for validation module. Should be removed from here.

```
Checks if the session key has been enabled, that it is not due and has not yet expired
Then passes the validation flow to appropriate Session Validation module
```

filmakarov: fixed thnx

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
filmakarov: fixed thnx

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
filmakarov: +

contracts/smart-account/modules/SessionKeyManagerModule.sol
filmakarov: This is the link to the discussion on this matter, I explain there the potential use case for such an event

https://biconomyworkspace.slack.com/archives/C05E6G3NTAN/p1691566571842799

please let me know wdyt

contracts/smart-account/modules/MultichainECDSAValidator.sol
filmakarov: we don't know , as `bytes32[] merkleProof` length can vary

test/bundler-integration/module/MultichainValidator.Module.specs.ts
```diff
@@ -0,0 +1,303 @@
...
+      verifyingPaymaster: await getVerifyingPaymaster(deployer, verifiedSigner),
+      multichainECDSAValidator: multichainECDSAValidator,
+      sessionKeyManager: sessionKeyManager,
+      sessionKeyMerkleTree: sessionKeyMerkleTree,
+    };
+  });
+
+  describe ("Multichain userOp validation", async () => {
+
+    it ("should process a userOp with a multichain signature", async () => {
```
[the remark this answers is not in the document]
filmakarov: 1) _add more test cases where you already have a merkle root (not generating for the first time) and you use cached leaf info to just update the tree_

I do not think it is need for the case when the tree is being build from userOps and not uploaded on-chain, but provided along with a userOp and signed. 
I do not think that dApps will store userOpHashes and userOps to then rebuild tree, I think this will be built, signed, and submitted on the go. Please refer to [this diagram](https://www.notion.so/biconomy/Multichain-Validator-module-Blog-Post-313272e5fe5d4ac88479a0ea788a5a44?pvs=4#7c099c8983b44161ab65425c8bc34a56)

2) For the session keys we have [such a test](https://github.com/bcnmy/scw-contracts/blob/adbffed2056806f9bc7a37b953ca663144a6160f/test/module/SessionKeyManager.Module.specs.ts#L103-L141). 

3) _how do we test multichain stuff using hardhat?_
We do not do it now.
```
 * @note Those tests do not actually process the userOp on several chains, instead they showcase
 * that one of the userOps included in the tree, which root is signed by user, can be processed
 * on the corresponding chain. Assuming, this will be valid for any userOp from the tree,
 * this approach can be considered representative enough for testing purposes.
 * The actual multichain tests based on Foundry framework can be added later
 ```
Can test it later on Foundry. For now I consider testing that any chain works is enough.

4) _can we add more comments at each step of test like it will enable session key on chain a and chain b and both will be validated by multichain validator? and what happens at each step._

added comments to the `test/module/MultichainValidator.Module.specs.ts`
For bundler integrated setupTests comments and flow are same that for regular.

test/bundler-integration/module/MultichainValidator.Module.specs.ts
filmakarov: yes. they are passed to cases from `setupTests`

filmakarov:
Factory is staked. One more issue (not on our end btw) detected.

Details: https://linear.app/biconomy/issue/SMA-1#comment-5eb1b60e

filmakarov:
Submitted changes as per discussion here:
https://linear.app/biconomy/issue/SMA-1/implement-multi-chain-signature-validator-module

Removed `extocodesize` check from `initForSmartAccount` method in `ECDSAOwnershipModule` 
- it won't be possible to deploy a SA with smart contract as owner via userOp even w/o this check
- for deployments via directly calling factory : 
-- via SDK: a check in the SDK can be implemented
-- directly:   `initForSmartAccount` parameter renamed `owner` => `eoaOwner` so it's cleat that eoa is expected to be there

filmakarov:
@livingrockrises you agree to merge?

filmakarov:
Merging as per https://linear.app/biconomy/issue/SMA-1#comment-e783a9ff

PR #103, opened 2023-08-03
feat: add time range in verifying paymaster + enable viaIR 

livingrockrises:
updated test cases (refer in pm service)

### DIFF
--- a/contracts/smart-contract-wallet/paymasters/verifying/singleton/VerifyingSingletonPaymaster.sol
+++ b/contracts/smart-contract-wallet/paymasters/verifying/singleton/VerifyingSingletonPaymaster.sol
@@ -6,7 +6,7 @@ pragma solidity 0.8.17;
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
 import {UserOperation, UserOperationLib} from "@account-abstraction/contracts/interfaces/UserOperation.sol";
-import {BasePaymaster, IEntryPoint} from "../../BasePaymaster.sol";
+import "../../BasePaymaster.sol";
 import {PaymasterHelpers, PaymasterData, PaymasterContext} from "../../PaymasterHelpers.sol";
 import {SingletonPaymasterErrors} from "../../../common/Errors.sol";
 
@@ -154,7 +154,9 @@ contract VerifyingSingletonPaymaster is
      */
     function getHash(
         UserOperation calldata userOp,
-        address paymasterId
+        address paymasterId,
+        uint48 validUntil,
+        uint48 validAfter
     ) public view returns (bytes32) {
         //can't use userOp.hash(), since it contains also the paymasterAndData itself.
         address sender = userOp.getSender();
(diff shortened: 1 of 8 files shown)


contracts/smart-contract-wallet/paymasters/verifying/singleton/VerifyingSingletonPaymaster.sol
```diff
@@ -6,7 +6,7 @@ pragma solidity 0.8.17;
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
 import {UserOperation, UserOperationLib} from "@account-abstraction/contracts/interfaces/UserOperation.sol";
-import {BasePaymaster, IEntryPoint} from "../../BasePaymaster.sol";
```
AmanRaj1608, line 9: No need to change the named imports i think

livingrockrises: I had to, otherwise it won't make use of Helpers.sol imported from BasePaymaster.

scripts/utils/index.ts
```diff
@@ -42,7 +42,7 @@ export enum DEPLOYMENT_SALTS {
   MULTI_SEND_CALLONLY = "DEVX_MULTI_SEND_CALLONLY_V0_11042023_pcnXVXc",
   WALLET_FACTORY = "DEVX_WALLET_FACTORY_V0_11042023_vyLkpGh",
   WALLET_IMP = "DEVX_WALLET_IMP_V0_11042023_AwPKF0R",
-  SINGELTON_PAYMASTER = "DEVX_SINGLETON_PAYMASTER_V0_11042023_v6ISI9i",
+  SINGELTON_PAYMASTER = "DEVX_SINGLETON_PAYMASTER_V1_03082023_0Af0vtw", // 0x0000064e9c653e373af18ef27f70be83df5476b7
```
tomarsachin2271, line 45: Shouldn't this be new field SINGLETON_PAYMASTER_V2?

What if we need to deploy v1 paymaster on other new chains? If not then on dashboard we have to consider showing only V2 as version for paymasters.

livingrockrises: It starts with 0 so I incremented it. on dashboard @divyaN73 suggested 1.1.0 which makes sense as it's not major update. besides you can change date or anything on the salt to deploy v1/any on other new chains

livingrockrises: also using this salt it's already deployed

filmakarov:
Can you please remind me, what have we decided regarding integrating paymasters into scw repo from the paymasters repo as a dependency?
are we going to integrate it into Modular branch?

tomarsachin2271 (approved):

livingrockrises:
> Can you please remind me, what have we decided regarding integrating paymasters into scw repo from the paymasters repo as a dependency? are we going to integrate it into the Modular branch?

All the paymasters code should be removed from scw repo and moved to paymasters repo and publish a package there. both repos can make use of published package from the other. Need to work on this and set up more structure in the paymasters repo.

PR #104, opened 2023-08-03
feat: Integrate Bundler into Hardhat Unit Tests

ankurdubey521:
# Introduction

UserOperations in the main public mempool are subject to restrictions including but not limited to what storage can be accessed, which opcodes can be called etc. These restrictions are applicable in the validation phase and are dependent upon what is being executed.

These restrictions are enforced by the bundler off-chain, and therefore it is important to test that the smart contracts are compliant with these restrictions. 

# Setup
The tests require an active `docker` runtime to be available on the system. Then, the [bundler-test.sh](https://github.com/bcnmy/scw-contracts/blob/test-bundler-integration/scripts/bundler-tests.sh) starts for starting an instance of geth in developer mode to act as the test-net and an instance of the infinitism bundler to submit user operations. Geth is needed because the default hardhat network does not have support for the `debug_traceCall` RPC method used by the bundler to enforce the required restrictions.

Once both services are up, the tests are run against the local test net. Since this is not a hardhat native network, we lose some of the utilities provided by hardhat like pre-funded addresses, the ability to rollback the chain to some point and many more. These two functionalities are implemented by the [BundlerTestEnvironment](https://github.com/bcnmy/scw-contracts/blob/test-bundler-integration/test/bundler-integration/environment/bundlerEnvironment.ts#L19) class.

The bundler and test net are shut down after all the tests have been executed.

## Current Progress
1. Implement the scripts required to set up the environment.
2. Add tests to verify that the bundler performs validations while submitting the userop.
3. Implement test for 'Ownerless Smart Account Basics' to [here](https://github.com/bcnmy/scw-contracts/blob/test-bundler-integration/test/bundler-integration/smart-account/SA.Basics.specs.ts#L22), with necessary modifications to submit via bundler.

If we're happy with this setup I can work on implementing the other tests.

# Usage
`yarn bundler-test`

# Sample Test Run
```
$ bash -c ./scripts/bundler-tests.sh
⚙️  1. Launching geth....
[+] Running 2/2
 ✔ Network environment_default       Created                                                                          0.1s
 ✔ Container environment-geth-dev-1  Started                                                                          0.4s
⚙️  2. Deploying Entrypoint...
No need to generate any newer typings.
Entrypoint deployed at 0x1E1EDc60c3C65E45C18D30a5D6E614c2e12d08E1
⚙️  3. Launching Bundler...
[+] Running 10/10
 ✔ bundler 9 layers [⣿⣿⣿⣿⣿⣿⣿⣿⣿]      0B/0B      Pulled                                                               29.2s
   ✔ 30bfbe1cefeb Pull complete                                                                                      11.5s
   ✔ 10bdaa199501 Pull complete                                                                                      11.7s
   ✔ ee9921bee3cc Pull complete                                                                                      22.7s
   ✔ 1468f2c2b412 Pull complete                                                                                      23.0s
   ✔ eb1c82bb32a7 Pull complete                                                                                      23.1s
   ✔ 8915e740b3d2 Pull complete                                                                                      23.1s
   ✔ c7b437feaf73 Pull complete                                                                                      24.0s
   ✔ b9451f25b420 Pull complete                                                                                      24.2s
   ✔ 20c2ef474774 Pull complete                                                                                      24.2s
[+] Running 1/1
 ✔ Container environment-bundler-1  Started                                                                           1.6s
⚙️  4. Waiting for Bundler to start...
Waiting for 200 OK response, got 000. Retrying in 5 seconds...
Received 200 OK response!
⚙️  5. Running tests...
No need to generate any newer typings.


  ✔ Default Signers should have funds after environment setup (1797ms)
  ✔ Should be able to revert to snapshot (2413ms)
  ✔ Should be able to submit user operation using bundler (1009ms)
  ✔ Should not submit user operation that calls TIMESTAMP in the validation phase (687ms)
  ✔ Can send an ERC20 Transfer userOp (1242ms)
  ✔ Can send a Native Token Transfer userOp (1363ms)
  1) Can send a userOp with Paymaster payment

  6 passing (11s)
  1 failing

  1) Can send a userOp with Paymaster payment:
     UserOperationSubmissionError: {"message":"paymaster uses banned opcode: BASEFEE","code":-32502}
      at BundlerTestEnvironment.sendUserOperation (test/bundler-integration/environment/bundlerEnvironment.ts:116:13)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Context.<anonymous> (test/bundler-integration/smart-account/SA.Basics.specs.ts:175:5)



⚙️  6. Stopping geth and bundler....
[+] Running 3/2
 ✔ Container environment-geth-dev-1  Removed                                                                          0.6s
 ✔ Container environment-bundler-1   Removed                                                                         10.3s
 ✔ Network environment_default       Removed                                                                          0.1s
✨  Done in 84.11s.
```


test/bundler-integration/environment/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: "2"
+
+services:
+  bundler:
+    ports: ["3000:3000"]
+    image: accountabstraction/bundler:0.6.1
+    command: --network http://geth-dev:8545 --entryPoint 0x1E1EDc60c3C65E45C18D30a5D6E614c2e12d08E1 --show-stack-traces
```
livingrockrises, line 7: which entry point address is this?
0x1E1EDc60c3C65E45C18D30a5D6E614c2e12d08E1

livingrockrises (commented):
reviewed. will run on my machine once before approving

test/bundler-integration/environment/docker-compose.yml
ankurdubey521: The address of the deployed entrypoint after calling https://github.com/bcnmy/scw-contracts/blob/test-bundler-integration/test/bundler-integration/environment/deployEntrypoint.ts, and the same as the the address of the entrypoint deployed in unit tests

livingrockrises: if it's not deterministic, is there some way we don't have to hardcode?

ankurdubey521: Yeah I just pushed a fix for that, testing it currently

filmakarov:
## Additional requirements
Required me to install `coreutils` via `brew install coreutils`. 
If your system already has `realpath` this step won't be required.

## How to use in tests
### Initiate environment and before/after hooks
```
let environment: BundlerTestEnvironment;
  let defaultSnapshot: Snapshot;

  before(async function () {
    const chainId = (await ethers.provider.getNetwork()).chainId;
    if (chainId !== BundlerTestEnvironment.BUNDLER_ENVIRONMENT_CHAIN_ID) {
      this.skip();
    }

    environment = await BundlerTestEnvironment.getDefaultInstance();
    defaultSnapshot = await environment.snapshot();
  });

  afterEach(async function () {
    const chainId = (await ethers.provider.getNetwork()).chainId;
    if (chainId !== BundlerTestEnvironment.BUNDLER_ENVIRONMENT_CHAIN_ID) {
      this.skip();
    }

    await environment.revert(defaultSnapshot);
  });
```

### Use ethers.getSigners()
Use `ethers.getSigners()` instead of `waffle.provider.getWallets();` to setup test EOAs. The latter complains that it can only be used with hardhat node, and throws with any other RPC.

### preVerificationGas
Manually set
```
{
        preVerificationGas: 50000,
      }
```
when creating the userOps.
The bundler has its own logic for estimating pre verification gas and usual estimate are around 44k. It complains if the gas is below that.

## Notes on bundler environment
Here are a few arguments for dividing the tests:
- With this environment, you lose a lot of the testing utilities that hardhat provides you, like the ability to set/skip block time, move backward/forward with blocks etc. For any tests that need these, we would need to re-implement these features through hardhat.
- The bundler returns the hash of the user-operation, but not the transaction hash of the actual handleOps call. This is an issue for any tests that tests events emitted in the call, as finding the actual tx hash in which the op was included could be tricky.
- Not all tests benefit from the bundler environment - Ideally, we should only be testing UserOperations that we know will not revert on-chain (happy flows). For everything which reverts, it’s more efficient to just use hardhat’s default node and test against the Entrypoint contract directly.
Lastly, the bundler environment is unstable right now, it may not be a good idea to completely switch to it

Thus, @ankurdubey521 agreed that the best way of using bundler env will be the following:

- Let's keep bundler-powered tests in a separate folder.
- Let's just test positive cases via bundler. Actually, this is enough as we just want to test that in a positive case bundler allows such operations
- Let's have this in mind when designing the test specs for the new modules and/or functionalities. So when we do this, we make the bundler-enabled specs to test all the positive test cases via bundler to make sure we don't use banned opcodes and do not access any restricted storage. And we make regular test specs to test negative cases / edge cases. To avoid duplicating the code, we can skip the cases that we already tested via bundler in specs which use regular hardhat environment

filmakarov (approved):
Approving. See the comment above for the notes on how to use bundler env

filmakarov:
Waiting for 
- @livingrockrises to run locally
- @ankurdubey521 to add grep support
- @ankurdubey521 can we also divide test cases by spec file here?
Screen Shot 2023-08-09 at 13 32 00
the same way hardhat test does it using `describe`s 

Then can merge imo

ankurdubey521:
# Changes:
1. Use a custom bundler image to incorporate eth-infinitism/bundler#133 fixes.
2. Divide test cases by spec file <details><summary>Sample Run</summary>Screenshot 2023-08-09 at 5 14 19 PM</details>
3. Allow passing arbitrary parameters to the hardhat test command. <details><summary>Sample Grep Usage</summary>Screenshot 2023-08-09 at 5 16 22 PM</details>
4. Implement:
    - SA.Modules.spec.ts
    - SA.Setup.specs.ts
    - SA.UserOps.specs.ts

ankurdubey521:
Usage in tests has also been tweaked to fix specific bugs:
```
  let deployer: SignerWithAddress,
    smartAccountOwner: SignerWithAddress,
    charlie: SignerWithAddress,
    verifiedSigner: SignerWithAddress;
  let environment: BundlerTestEnvironment;

  before(async function () {
    const chainId = (await ethers.provider.getNetwork()).chainId;
    if (chainId !== BundlerTestEnvironment.BUNDLER_ENVIRONMENT_CHAIN_ID) {
      this.skip();
    }

    environment = await BundlerTestEnvironment.getDefaultInstance();
  });

  beforeEach(async function () {
    [deployer, smartAccountOwner, charlie, verifiedSigner] =
      await ethers.getSigners();
  });

  afterEach(async function () {
    const chainId = (await ethers.provider.getNetwork()).chainId;
    if (chainId !== BundlerTestEnvironment.BUNDLER_ENVIRONMENT_CHAIN_ID) {
      this.skip();
    }

    await Promise.all([
      environment.revert(environment.defaultSnapshot!),
      environment.resetBundler(),
    ]);
  });```
  
 The default snapshot is stored in the `BundlerEnvironment` object automatically.

filmakarov:
> Usage in tests has also been tweaked to fix specific bugs:

Could you please note, what has changed? 
Now every wallet is declared separately?

ankurdubey521:
@filmakarov Well, earlier I was calling `await ethers.getSigner()` at the time of declaration of the signers. However, this is not correct as all async initialisations must happen in a hook (`before`, `beforeEach`), and this was the reason why test cases were not being grouped by describe blocks in the test run.

Consequently, all signers must be declared separately.

livingrockrises:
brew install coreutils can be part of the script or goes in README.

livingrockrises:
should I update the paymaster contract here? or we can ignore the failing test. merging this PR is good

livingrockrises (approved):
LGTM

ankurdubey521:
> should I update the paymaster contract here? or we can ignore the failing test. merging this PR is good

A separate PR can address the failing tests for
1. Paymaster
2. Owner Registry Module
3. V2 to V3 Upgrade Tests

filmakarov:
I will handle 2 and 3.

@livingrockrises can we integrate paymaster as dependency from the paymasters repo?

livingrockrises:
> I will handle 2 and 3.
> 
> @livingrockrises can we integrate paymaster as dependency from the paymasters repo?

yes let's create a task for that. will pick it up or delegate to someone soon.

PR #105, opened 2023-08-04
remove gasPrice from context and in postOp use tx.gasprice

livingrockrises:
userop.gasprice() accesses block.basefee and adds priority fee to it

in validateUserOp it should not be used. if gas price is needed use maxFeePerGas (its not needed here)

but in postop it's needed where postop could either use maxFeePerGas and maxPriorityFeePerGas passed from the context and now it can use block.basefee. 

or don't pass anything in context and just use tx.gasprice which will be effective gasprice for any chain. [ This approach is used here , also already used in token paymaster ]


contracts/smart-contract-wallet/paymasters/verifying/singleton/VerifyingSingletonPaymaster.sol
```diff
@@ -248,9 +252,10 @@ contract VerifyingSingletonPaymaster is
     ) internal virtual override {
         PaymasterContext memory data = context._decodePaymasterContext();
         address extractedPaymasterId = data.paymasterId;
+        uint256 effectiveGasPrice = block.basefee + data.maxPriorityFeePerGas;
```
tomarsachin2271, line 255: We should use 
uint256 effectiveGasPrice = min(data.maxFeePerGas, data.maxPriorityFeePerGas + block.basefee)

function min(uint256 a, uint256 b) internal pure returns (uint256) {
        return a < b ? a : b;
    }

livingrockrises: done

tomarsachin2271 (approved):

ankurdubey521 (approved):
Confirmed locally, the new implementation's user operations are able to be submitted by the reference bundler.

PR #106, opened 2023-08-05
fix: YULException for Stack too Deep Error while compiling with viaIR: true

ankurdubey521:
This PR fixes the error in ModuleManager.sol while compiling with `viaIR` enabled.
```
YulException: Cannot swap Variable var_to_offset with Variable var_success: too deep in the stack by 3 slots in [ var_to_offset RET var_to_offset var_to_length var_data_offset var_value_offset var_value_length var_operations_offset var_operations_length var_data_length var_i var_i var_to_length var_data_offset var_value_offset var_value_length var_operations_offset var_operations_length var_data_length var_success ]
No memoryguard was present. Consider using memory-safe assembly only and annotating it via 'assembly ("memory-safe") { ... }'.


Error HH600: Compilation failed

For more info go to https://hardhat.org/HH600 or run Hardhat with --show-stack-traces
```

### DIFF
--- a/contracts/smart-account/base/ModuleManager.sol
+++ b/contracts/smart-account/base/ModuleManager.sol
@@ -238,27 +238,29 @@ abstract contract ModuleManager is
 
         for (uint256 i; i < to.length; ) {
             // Execute transaction without further confirmations.
-            success = execute(
+            success = _executeFromModule(
                 to[i],
                 value[i],
                 data[i],
-                operations[i],
-                gasleft()
+                operations[i]
             );
-            if (success) {
-                emit ModuleTransaction(
-                    msg.sender,
-                    to[i],
-                    value[i],
-                    data[i],
-                    operations[i]
-                );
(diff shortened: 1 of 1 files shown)


ankurdubey521:
Interestingly, annotating the assembly block with "memory-safe" in ModuleManager.sol and Executor.sol did not fix the issue. Annotating all assembly blocks in the codebase fixes the issue, but I didn't spend time going through every assembly block the codebase to verify if they were memory safe, so I opted for this solution instead.

filmakarov (approved):
I think this can be merged safely to the audited [SCW-V2-Modular-SA](https://github.com/bcnmy/scw-contracts/tree/SCW-V2-Modular-SA) branch, as it just refactors the code. 
So it doesn't require additional checks from auditors.

livingrockrises:
I remember now I had also got this issue and adding "memory-safe" to assembly blocks had it fixed for me.

livingrockrises (approved):
LGTM

PR #108, opened 2023-08-11
Stakeable Factory

filmakarov:
Bundler tests show, that we need to stake our SmartAccountFactory as it uses storage associated with the sender.

[ERC-4337](https://eips.ethereum.org/EIPS/eip-4337#un-staked-entities): "If the UserOp doesn’t create a new account (that is initCode is empty), or the UserOp creates a new account using a staked factory contract, then the entity may also use [storage associated with the sender](https://eips.ethereum.org/EIPS/eip-4337#storage-associated-with-an-address))".

To stake the factory, we need to implement `addStake`, `unlockStake`, `withdrawStake(address payable withdrawAddress)` methods.
The last two should be access-protected. i.e. ownable.
So our SmartAccountFactory will become Stakeable => Ownable. However, `onlyOwner` modifier will be applicable only to abovementioned 3 methods.

This PR introduces Stakeable.sol (applicable not only to the Factory but to any entity) and makes Factory stakeable via inheriting this class.


test/Smart-Account-Factory.specs.ts
```diff
@@ -20,40 +21,68 @@ describe("Smart Account Factory", async () => {
...
+  describe("Stakeable", async () => {
+    it ("can add stake to the EP", async () => {
+      const { smartAccountFactory, smartAccountImplementation, entryPoint } = await setupTests();
+      const stakeAmount = ethers.utils.parseEther("1.234256");
+      await smartAccountFactory.addStake(entryPoint.address, 600, { value: stakeAmount });
+      const depositInfo = await entryPoint.getDepositInfo(smartAccountFactory.address);
+      expect(depositInfo.stake).to.equal(stakeAmount);
+    });
+
+    // can unlock
```
tomarsachin2271, line 80: Unimplemented test cases. Should implement these test cases as well

contracts/smart-account/common/Stakeable.sol
```diff
@@ -0,0 +1,33 @@
...
+contract Stakeable is Ownable {
+    constructor(address _newOwner) {
+        _transferOwnership(_newOwner);
+    }
+
+    function addStake(
+        address epAddress,
+        uint32 unstakeDelaySec
+    ) external payable onlyOwner {
+        IEntryPoint(epAddress).addStake{value: msg.value}(unstakeDelaySec);
```
ankurdubey521, line 20: What's your opinion on zero address checks? I know some auditors flag any functions that do not have zero address checks.

test/Smart-Account-Factory.specs.ts
filmakarov: Yes, I will pass them to Darpit.

contracts/smart-account/common/Stakeable.sol
filmakarov: I think it's better to do zero address checks, however, those methods are ownable, so not public. I don't think owner will pass zero_address.
However, since the methods won't be called often, I can add zero address check despite they spend gas

test/Smart-Account-Factory.specs.ts
filmakarov: I suggest merging in without those tests, as those methods just forward everything to the EP. 
Until it is merged, I can not stake the factory to test Multichain Validator via bundler properly.

ankurdubey521 (approved):

livingrockrises (approved):
LGTM

PR #110, opened 2023-08-16
lint-fix scripts

livingrockrises:


ankurdubey521:
Would it be possible to configure this to run before very commit, similar to the solidity formatter?

PR #115, opened 2023-08-23
experimental: Linux Compatibility for Bundler Integration Tests

ankurdubey521:
In the Linux build of geth's docker image, the implementation of `debug_setHead` appears to be broken - the node immediately crashes after calling this RPC method. This is not the case with their M1 builds, where the RPC works fine.

We use this RPC method to reset the state of the blockchain after each test case is run, to ensure that each test is running under the same conditions. This PR removes the usage of this RPC, and introduces "hacks" to account for the fact that the blockchain state between each test is shared.
These hacks include:
1. Randomizing the `smartAccountOwner` address between each test run to ensure that the smart account is deployed to a new address for each test, otherwise the test fails with `CREATE2 failed`
2. Turning on `skipIfAlreadyDeployed` in the deployment script for `Entrypoint`.

In general, this results in a degraded testing experience and results in a significant increase in complexity while writing tests. I have created an issue on geth to hopefully get this fixed on their end [here](https://github.com/ethereum/go-ethereum/issues/27990).

### DIFF
--- a/src/deploy/01_deploy_entrypoint.ts
+++ b/src/deploy/01_deploy_entrypoint.ts
@@ -11,6 +11,7 @@ const deploy: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     args: [],
     log: true,
     deterministicDeployment: true, // Needed for bundler tests to ensure the entrypoint address does not change b/w tests
+    skipIfAlreadyDeployed: true,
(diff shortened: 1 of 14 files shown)


ankurdubey521:
This bug has been fixed by geth, therefore closing this PR

PR #116, opened 2023-08-24
Add explicit selector check to ERC20SessionValidationModule

filmakarov:
Add explicit `execute` and `execute_ncC` selector check in order to avoid validating calls to other Smart Account methods through this module
```
// execute(address,uint256,bytes)
bytes4 public constant EXECUTE_SELECTOR = 0xb61d27f6;

// execute_ncC(address,uint256,bytes)
bytes4 public constant EXECUTE_OPTIMIZED_SELECTOR = 0x0000189a;
```


livingrockrises:
you can also keep MaliciousSessionKeyValidationModule (just like Mock) one and write negative test case to update implementation. so we can reproduce and confirm other cases such ways

contracts/smart-account/modules/SessionValidationModules/ISessionValidationModule.sol
```diff
@@ -3,7 +3,10 @@ pragma solidity 0.8.17;
 
 import {UserOperation} from "../BaseAuthorizationModule.sol";
 
-interface ISessionValidationModule {
+abstract contract ISessionValidationModule {
+    bytes4 public constant EXECUTE_SELECTOR = 0xb61d27f6;
```
livingrockrises, line 7: add the comments here that method name is execute() and optimised execute is excute_xxxxx

livingrockrises (commented):
reviewed

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
```diff
@@ -28,7 +28,13 @@ contract ERC20SessionValidationModule {
         bytes32 _userOpHash,
         bytes calldata _sessionKeyData,
         bytes calldata _sessionKeySignature
-    ) external pure returns (bool) {
+    ) external pure override returns (bool) {
+        require(
+            bytes4(_op.callData[0:4]) == EXECUTE_SELECTOR ||
```
ankurdubey521, line 33: I think we should start using custom errors in Solidity. It would help reduce deployment costs and be more actionable for the client since they can be parameterized.
https://soliditylang.org/blog/2021/04/21/custom-errors/

ankurdubey521: also, is `execute` expected to be called more often than `execute_xxx`? If not let's switch the order in which the selector is compared so that the more common case is evaluated first.

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
```diff
@@ -28,7 +28,13 @@ contract ERC20SessionValidationModule {
...
+                bytes4(_op.callData[0:4]) == EXECUTE_OPTIMIZED_SELECTOR,
+            "ERC20SV Invalid Selector"
+        );
+
```
ankurdubey521, line 37: Potential optimisation on line 54: != instead of >
https://twitter.com/filmakarov/status/1638540351815254016 😅

test/module/SessionValidationModules/ERC20SessionValidation.Module.specs.ts
```diff
@@ -225,5 +225,82 @@ describe("SessionKey: ERC20 Session Validation Module", async () => {
...
+        callData: spoofedData, //use spoofed data
+        preVerificationGas: 50000,
+      },
+      sessionKey,
+      entryPoint,
+      'nonce'
+    );
+
+    const paddedSig = ethers.utils.defaultAbiCoder.encode(
+      //validUntil, validAfter, sessionVerificationModule address, validationData, merkleProof, signature
```
ankurdubey521, line 273: Unsure about leaving this commented out

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
filmakarov: 1) yep, we do use custom errors and it was actually my suggestion :) however, in this specific case the revert message is appended to AA23 message by EP
```
revert FailedOp(opIndex, string.concat("AA23 reverted: ", revertReason));
```
so no reason to use custom error here.

2) fair note, will do

contracts/smart-account/modules/SessionValidationModules/ISessionValidationModule.sol
```diff
@@ -18,5 +21,5 @@ interface ISessionValidationModule {
         bytes32 _userOpHash,
         bytes calldata _sessionKeyData,
         bytes calldata _sessionKeySignature
-    ) external view returns (bool);
+    ) external view virtual returns (bool);
```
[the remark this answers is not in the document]
filmakarov: I would leave it as `view` or even not `view` so 3rd parties modules can be more flexible in what they do in this method.

contracts/smart-account/modules/SessionValidationModules/ISessionValidationModule.sol
filmakarov: +

contracts/smart-account/modules/SessionValidationModules/ERC20SessionValidationModule.sol
filmakarov: done

livingrockrises (approved):

ankurdubey521 (approved):

PR #117, opened 2023-08-25
2D nonce feature 

livingrockrises:
https://www.notion.so/biconomy/2D-nonces-in-SA-7523daf610114fd7a6d4d4cf147fe015

decided to go with having nonce(uint198 _key) only one nonce related method which is in line with entry point. test cases are upgrade accordingly (can be improved!) to use with v1 and v2 accounts.


test/utils/userOp.ts
```diff
@@ -335,7 +362,10 @@ export async function makeEcdsaModuleUserOp(
     },
     userOpSigner,
     entryPoint,
-    "nonce"
+    "getNonce",
```
filmakarov, line 365: Could you please explain, why we use `getNonce` here while we don't have `getNonce` method in SA?

test/bundler-integration/smart-account/SA.concurrent.txn.specs.ts
```diff
@@ -0,0 +1,175 @@
...
+        encodeTransfer(charlie.address, tokenAmountToTransfer.toString()),
+      ],
+      userSA.address,
+      smartAccountOwner,
+      entryPoint,
+      ecdsaModule.address,
+      {
+        preVerificationGas: 50000,
+      },
+      5 // nonceKey
```
filmakarov, line 133: I would suggest making 
```
const erc20TransferUserOpNonceKey = 5;
```
instead of adding comment. and same for the second userOp

filmakarov (approved):
I only have one question and a minor comment.
Approved.

test/bundler-integration/smart-account/SA.concurrent.txn.specs.ts
livingrockrises: done

test/utils/userOp.ts
livingrockrises: I can leave it as "nonce". It doesn't matter when flag below (useNonceKey) is sent true - it then always uses nonce(uint198) method. passed nonce function name is used if the mentioned flag is false.

livingrockrises: updated everywhere. all test cases still pass as expected

ankurdubey521 (approved):

PR #122, opened 2023-08-31
Features/sma 115 package json

Aboudjem:
removed unused packages
removed package-lock.json
removed `git add .` from husky hook


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@biconomy/account-contracts",
+  "name": "@bcnmy/scw-contracts",
```
tomarsachin2271, line 2: Don't think we should change it right now.. need some inputs from @livingrockrises here.

tomarsachin2271 (commented):
Left a comment. Other changes looks good.

package.json
livingrockrises: we don't have any organisation on npm named 'bcnmy'. If anything we can rename on github from bcnmy -> biconomy. but on npm imo these should start with @biconomy... only

Aboudjem: I named it like the repo but you're right. Should we use the same name everywhere to have better consistency? How about adding `account-abstraction` to the name? Like `biconomy-account-abstraction`. What do you think?

tomarsachin2271: For better discoverability we can use
@biconomy/account-abstraction

livingrockrises: no. we are talking about scw contracts package name or organisation name?

livingrockrises: Adam suggested org name and @tomarsachin2271 you suggestion is mix up for a package name for account contracts. @biconomy/smart-account or @biconomy/aa-smart-account will also have good visibility.

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm run lint:fix && git add .
+npm run lint:fix
```
livingrockrises, line 4: why remove git add?

package.json
```diff
@@ -73,39 +82,11 @@
     "typescript": "^4.7.3",
```
livingrockrises, line 82: can we also update typechain versions and prepack/build scripts

livingrockrises (commented):
reviewed

.husky/pre-commit
Aboudjem: Removing `git add .` helps us to choose the specific files we want for commit and it stops us from adding files we don't want by mistake, automatically adding all changed files to the commit can be somehow  dangerous as you might unintentionally commit files that you didn't intend to

package.json
Aboudjem: I will keep it unchanged for now, we'll discuss it later

PR #141, opened 2023-10-03
SMA-153,SMA-150: Refactor

ankurdubey521:
# Summary

1. Import from OpenZeppelin instead of custom contracts wherever possible.
2. Move natspec, structs, error and events to dedicated interfaces.
3. Remove all Paymaster Contracts and Tests.

Related Issue: # (issue number)

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist


Aboudjem:
Very good ! just two things:

- as mentioned during the DSM, `fixes/` is mostly when we need to fix something in prod, otherwise `features/` should be used, even though refactoring is not a feature 😅
- I think it would be better to have each interface inside their respective folders, e.g `folder/File.sol` and `folder/IFile.sol` or `folder/File.sol` and `folder/interface/IFile.sol`

ankurdubey521:
> 2. It would be more organized to have each interface located in its respective folder. For example, `folder/File.sol` and `folder/IFile.sol`, or `folder/File.sol` and `folder/interface/IFile.sol`.
I see what you're saying, but for an external developer looking for just the interfaces, I think having it in one place makes more sense. Here's an example from uniswap's repo: https://github.com/Uniswap/v3-core/tree/main/contracts/interfaces

Aboudjem:
> > 2. It would be more organized to have each interface located in its respective folder. For example, `folder/File.sol` and `folder/IFile.sol`, or `folder/File.sol` and `folder/interface/IFile.sol`.
> 
> I see what you're saying, but for an external developer looking for just the interfaces, I think having it in one place makes more sense. Here's an example from uniswap's repo: https://github.com/Uniswap/v3-core/tree/main/contracts/interfaces

Agree that's a good suggestion, in this case let's follow uniswap and move all the interfaces (modules included) into the interfaces folder and create sub-folders

Aboudjem (approved):

AmanRaj1608 (approved):

PR #148, opened 2023-10-15
Social Recovery Module

filmakarov:
# Summary

Account Recovery Module for Biconomy Modular Smart Account.
Based on [this](https://vitalik.ca/general/2021/01/11/recovery.html).

## Features
- It allows the submission and execution of recovery requests
- EOA guardians only
- For security reasons, the guardian's address is not stored, instead, its signature over CONTROL_HASH is used
- Security delay is always applied to new guardians and recovery requests

## How it works
The algorithm relies on the 4337 concept of having the validation and execution phase.
The checks happen in the `validateUserOp` method.
Screenshot 2023-10-16 at 11 44 53

[Related Issue](https://linear.app/biconomy/issue/SMA-80/account-recovery-module-development)

## Change Type: New Feature

# Checklist


livingrockrises:
How do we check if userOp.calldata matches previously submitted recovery request?

How are guardian signatures supplied?

livingrockrises:
The above diagram tries to explain what happens during validation but also has mentions of execution. Can we have separate flow diagram on what happens during the execution?

Do we have a technical draft on how social recovery module works? like we have done for other modules

.solhintignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 artifacts
 contracts/smart-account/test
-contracts/smart-account/libs
\ No newline at end of file
+contracts/smart-account/libs
+contracts/smart-account/modules/AccountRecoveryModule.sol
```
livingrockrises, line 5: we plan to fix linting soon for these contracts? in different PR?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,580 @@
...
+    struct SaSettings {
+        uint8 guardiansCount;
+        uint8 recoveryThreshold;
+        uint48 securityDelay;
+    }
+
+    struct RecoveryRequest {
+        bytes32 callDataHash;
+        uint48 requestTimestamp;
+    }
```
Aboudjem, line 44: Could we move this to IAccountRecoveryModule.sol ?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,580 @@
...
+
+    string public constant NAME = "Account Recovery Module";
+    string public constant VERSION = "0.1.0";
+
+    bytes32 public constant CONTROL_HASH =
```
livingrockrises, line 49: I would prefer to store it like 0x33e4b41198cc5b8053630ed667ea7c0c4c873f7fc8d9a478b5d7259cec0a4a00 directly what kecckak value it is. and mention in comment above like it is actually calculated like this...

// keccak256(abi.encodePacked("ACCOUNT RECOVERY GUARDIAN SECURE MESSAGE"));

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,580 @@
...
+        keccak256(abi.encodePacked("ACCOUNT RECOVERY GUARDIAN SECURE MESSAGE"));
```
Aboudjem, line 50: can we replace keccak256 + abi.encodePacked with Precomputed Hash?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,580 @@
...
+    error ThresholdTooHigh(uint8 threshold, uint256 guardiansExist);
+    error ZeroThreshold();
+    error InvalidAmountOfGuardianParams();
+    error GuardiansAreIdentical();
+    error GuardianNotExpired(bytes32 guardian, address smartAccount);
+    error EmptyRecoveryCallData();
+    error RecoveryRequestAlreadyExists(
+        address smartAccount,
+        bytes32 requestCallDataHash
+    );
```
Aboudjem, line 100: Could you move this as well to the interface?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,580 @@
...
+            if (guardians[i] == bytes32(0)) revert ZeroGuardian();
+            if (timeFrames[i].validUntil == 0)
+                timeFrames[i].validUntil = type(uint48).max;
+            if (timeFrames[i].validUntil < timeFrames[i].validAfter)
+                revert InvalidTimeFrame(
+                    timeFrames[i].validUntil,
+                    timeFrames[i].validAfter
+                );
+            if (
+                timeFrames[i].validUntil != 0 &&
```
Aboudjem, line 141: `timeFrames[i].validUntil != 0` might have a redundant condition since `timeFrames[i].validUntil` is set to `type(uint48).max`

Aboudjem (changes requested):

livingrockrises (commented):
commented (i)

contracts/smart-account/modules/AccountRecoveryModule.sol
ankurdubey521: Afaik this should not be a problem gas wise since constants are resolved at compile time.

ankurdubey521:
Just to confirm, should this be merged into `main` or `develop`?

filmakarov:
> How do we check if userOp.calldata matches previously submitted recovery request?
```
if (
            keccak256(userOp.callData) ==
            _smartAccountRequests[msg.sender].callDataHash
        )
```

> How are guardian signatures supplied?

they are in the userOp.signature field

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: Thank you! 🙏

.solhintignore
filmakarov: >AccountRecoveryModule.sol
it was added here because it violates the rule about function max length in lines and cyclomatic complexity.
All other linting is in place.
`validateUserOp` is longer than expected, but for the sake of gas savings decided to keep it as is. Check [this](https://biconomyworkspace.slack.com/archives/C05E6G3NTAN/p1697098546139319) thread for more details

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: fixed

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: Yep, agree with Ankur.
However, since Adam and Chirag both mentioned this, I pasted the precomputed hash

filmakarov:
> The above diagram tries to explain what happens during validation but also has mentions of execution. Can we have separate flow diagram on what happens during the execution?

Updated the diagram

> Do we have a technical draft on how social recovery module works? like we have done for other modules

Not yet. Will start making it this week

livingrockrises:
> > The above diagram tries to explain what happens during validation but also has mentions of execution. Can we have separate flow diagram on what happens during the execution?
> 
> Updated the diagram
> 
> > Do we have a technical draft on how social recovery module works? like we have done for other modules
> 
> Not yet. Will start making it this week

cool

for devs to understand what would help is ways to interact with this module. which would translate to different types of transaction journeys, for each can we add small flow like EP calls SA calls ARM and so on.

livingrockrises:
can we please add test cases for adding, removing and replacing guardians. + removing expired guardian

livingrockrises:
> > > The above diagram tries to explain what happens during validation but also has mentions of execution. Can we have separate flow diagram on what happens during the execution?
> > 
> > 
> > Updated the diagram
> > > Do we have a technical draft on how social recovery module works? like we have done for other modules
> > 
> > 
> > Not yet. Will start making it this week
> 
> cool
> 
> for devs to understand what would help is ways to interact with this module. which would translate to different types of transaction journeys, for each can we add small flow like EP calls SA calls ARM and so on.

like 

1. When account is deployed by default it enables social recovery module from the sdk. optionally setup can also be done at this step if dev/user provides at least one guardian. secure frontends of wallet onboarding would work like this 

btw ARM can not only be sole module enabled right? as it will only use validateUserOp for the cases when there is specific userop calldata

2. transaction journey to setup guardian on enabled module

3. flow for submit recovery request. who can submit, which details are required etc in calldata and signature and particular order to follow in stitching signatures.

4. flow for execute recovery request. who can submit, which details are required etc

Aboudjem:
please change the scope to merge into develop branch

filmakarov:
> please change the scope to merge into develop branch

done

filmakarov:
> can we please add test cases for adding, removing and replacing guardians. + removing expired guardian

they are in place.
https://github.com/bcnmy/scw-contracts/blob/2d016552d59dfd174b29af0b688a4771a44d5c4c/test/module/AccountRecovery.Module.specs.ts#L2072-L2129

https://github.com/bcnmy/scw-contracts/blob/2d016552d59dfd174b29af0b688a4771a44d5c4c/test/module/AccountRecovery.Module.specs.ts#L2277-L2357

etc

filmakarov:
> > > > The above diagram tries to explain what happens during validation but also has mentions of execution. Can we have separate flow diagram on what happens during the execution?
> > > 
> > > 
> > > Updated the diagram
> > > > Do we have a technical draft on how social recovery module works? like we have done for other modules
> > > 
> > > 
> > > Not yet. Will start making it this week
> > 
> > 
> > cool
> > for devs to understand what would help is ways to interact with this module. which would translate to different types of transaction journeys, for each can we add small flow like EP calls SA calls ARM and so on.
> 
> like
> 
> 1. When account is deployed by default it enables social recovery module from the sdk. optionally setup can also be done at this step if dev/user provides at least one guardian. secure frontends of wallet onboarding would work like this
> 
> btw ARM can not only be sole module enabled right? as it will only use validateUserOp for the cases when there is specific userop calldata
> 
> 2. transaction journey to setup guardian on enabled module
> 3. flow for submit recovery request. who can submit, which details are required etc in calldata and signature and particular order to follow in stitching signatures.
> 4. flow for execute recovery request. who can submit, which details are required etc

Yep, thank you! Will make all of those flowcharts in the article

livingrockrises:
do resolve the conflicts as well

Aboudjem (approved):

contracts/smart-account/interfaces/IAccountRecoveryModule.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+interface IAccountRecoveryModule {
```
ankurdubey521, line 4: what are your thoughts about declaring all public and external functions in the interface itself?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,528 @@
...
+     * @param guardians the list of guardians
+     * @param timeFrames validity timeframes for guardians
+     * @param recoveryThreshold how many guardians' signatures are required to authorize recovery request
+     * @param securityDelay amount of time required to pass between the submission of the recovery request
+     * and its execution
+     * @dev no need for explicit check `length == 0` as it is covered by `recoveryThreshold > length` and
+     * `recoveryThreshold == 0` cheks. So length can never be 0 while recoveryThreshold is not 0
+     */
+    function initForSmartAccount(
+        bytes32[] memory guardians,
```
ankurdubey521, line 64: is it cheaper to `memory` arrays here rather than `calldata`?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,528 @@
...
+            revert ThresholdTooHigh(recoveryThreshold, length);
+        if (recoveryThreshold == 0) revert ZeroThreshold();
+        if (length != timeFrames.length) revert InvalidAmountOfGuardianParams();
+        _smartAccountSettings[msg.sender] = SaSettings(
+            uint8(length),
+            recoveryThreshold,
+            securityDelay
+        );
+        for (uint256 i; i < length; ) {
+            if (guardians[i] == bytes32(0)) revert ZeroGuardian();
```
ankurdubey521, line 82: what happens if there are duplicate guardians in `_guardians`?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,528 @@
...
+            emit GuardianAdded(msg.sender, guardians[i], timeFrames[i]);
+            unchecked {
+                ++i;
+            }
+        }
+        return address(this);
+    }
+
+    /**
+     * @dev validates userOps to submut and execute recovery requests
```
ankurdubey521, line 103: typo in `submut`

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,528 @@
...
+
+        // otherwise we need to check all the signatures first
+        uint256 requiredSignatures = _smartAccountSettings[userOp.sender]
+            .recoveryThreshold;
+        if (requiredSignatures == 0) revert("AccRecovery: Threshold not set");
+
+        bytes calldata moduleSignature = userOp.signature[96:];
+
+        require(
+            moduleSignature.length >= requiredSignatures * 2 * 65,
```
ankurdubey521, line 142: A comment about these magic numbers would be great

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,528 @@
...
+        for (uint256 i; i < requiredSignatures; ) {
+            address currentUserOpSignerAddress = userOpHash
+                .toEthSignedMessageHash()
+                .recover(moduleSignature[2 * i * 65:(2 * i + 1) * 65]);
+
+            currentGuardianSig = moduleSignature[(2 * i + 1) * 65:(2 * i + 2) *
+                65];
+
+            currentGuardianAddress = CONTROL_HASH
+                .toEthSignedMessageHash()
```
ankurdubey521, line 163: It looks like `CONTROL_HASH.toEthSignedMessageHash()` is being calculated again in every iteration of the loop, maybe we should cache this value outside the loop or store it instead of `CONTROL_HASH` as a `bytes immutable`.

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,528 @@
...
+        address lastGuardianAddress;
+        address currentGuardianAddress;
+        bytes memory currentGuardianSig;
+        uint48 validAfter;
+        uint48 validUntil;
+        uint48 latestValidAfter;
+        uint48 earliestValidUntil = type(uint48).max;
+
+        for (uint256 i; i < requiredSignatures; ) {
+            address currentUserOpSignerAddress = userOpHash
```
ankurdubey521, line 155: `userOpHash.toEthSignedMessageHash()` can be cached outside the loop

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -0,0 +1,528 @@
...
+
+            currentGuardianAddress = CONTROL_HASH
+                .toEthSignedMessageHash()
+                .recover(currentGuardianSig);
+
+            if (currentUserOpSignerAddress != currentGuardianAddress) {
+                return SIG_VALIDATION_FAILED;
+            }
+
+            validAfter = _guardians[keccak256(currentGuardianSig)][
```
ankurdubey521, line 170: possible (minor) optimisation: calculate `keccak256(currentGuardianSig)` only once

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: `TimeFrame[]` can't be `calldata` here, but for `guardians` you're right, it will be probably cheaper to make it `calldata`

ankurdubey521: > TimeFrame[] can't be calldata here
Why is this the case?

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: Fair point! This could lead to unexpected behavior. Fixed and added proper test

filmakarov: 🙏

contracts/smart-account/interfaces/IAccountRecoveryModule.sol
filmakarov: I don't know if this is needed in this exact case. I don't expect people interact a lot with our Account Recovery through the interface or inherit the interface in other contracts.
wdyt?

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: fixed thnx

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: fair, storing it and not recalculating it saves 28 gas on average.

contracts/smart-account/interfaces/IAccountRecoveryModule.sol
ankurdubey521: yeah, I don't think it's necessary but it's just another good-to-follow practice. I did this for all other contracts for the last refactor I merged in develop

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: same for userOpHash

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: yep, made it

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: saves 44 gas on average, made it

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: +

ankurdubey521 (approved):
LGTM, just need to resolve a merge conflict

contracts/smart-account/interfaces/IAccountRecoveryModule.sol
filmakarov: made it to preserve the consistency 🤝

ankurdubey521: Thank you!

PR #152, opened 2023-10-22
fixes/passkey registry module init method

mohamadhammoud:
# Summary

Savings on Storage Reads: This code reduces the number of times data is read from cold storage, which saves gas. In the original code, storage was accessed multiple times, but this code accesses it only once.

Code Clarity: The refactored code is easier to understand and maintain, which can prevent bugs and improve the code's long-term quality.

## Change Type

- [ ] Bug Fix
- [X] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

- [X] My code follows this project's style guidelines
- [X] I've reviewed my own code
- [X] My changes generate no new warnings
- [X] All unit tests pass locally with my changes


Aboudjem:
Hey Mohamad, big thanks for your ongoing contributions, especially on gas saving this time! Do you have any metrics to share on how much gas we could save with this change?

contracts/smart-account/modules/PasskeyRegistryModule.sol
```diff
@@ -33,17 +33,17 @@ contract PasskeyRegistryModule is
         uint256 _pubKeyY,
         string calldata _keyId
     ) external override returns (address) {
-        if (
-            smartAccountPassKeys[msg.sender].pubKeyX != 0 &&
-            smartAccountPassKeys[msg.sender].pubKeyY != 0
-        ) {
+        PassKeyId memory passKeyId = smartAccountPassKeys[msg.sender];
```
ankurdubey521, line 36: could you please share a gas comparison b/w `PassKeyId memory passKeyId` and `PassKeyId storage passKeyId` ?

mohamadhammoud: I tried both gas comparison b/w **PassKeyId memory passKeyId** and **PassKeyId storage passKeyId**, 
**PassKeyId memory passKeyId** costs more gas even from the main code in the repo.
**PassKeyId storage passKeyId** costs less from the main code and the memory code.
Screenshot 2023-10-22

After trying the three approaches, I analysed that the **PassKeyId storage passKeyId** is more gas optimised, but I incorrectly submitted the **PassKeyId memory passKeyId**

Here I share the difference between the main repo code, and the last submitted code I pushed
the down table gas optimisation is  for the main repo, and the upper is the **PassKeyId storage passKeyId** one

ankurdubey521: That makes sense, given that with a memory declaration you're forcing the object to be copied to memory when it's not really need.

With the storage pointer, you're essentially saving one computation for calculating the slot at which object is stored, compared to the original implementation.

mohamadhammoud: That was the only reason of this PR.

Thank you.

Aboudjem (approved):

PR #158, opened 2023-10-25
Feat/delegate call module

livingrockrises:
# Summary

module has external method execTransaction which can be called (usually by smart account). 
if not called by smart account it won't work because we call execTransactionFromModule on the msg.sender
if module is not enabled on SA then execTransactionFromModule would not go through.

this is an example of simple execution module which opens up possibility to do delegate calls

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

# Additional Information

# Branch Naming

### DIFF
--- a/contracts/smart-account/modules/SimpleExecutionModule.sol
+++ b/contracts/smart-account/modules/SimpleExecutionModule.sol
@@ -1,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {ReentrancyGuard} from "../common/ReentrancyGuard.sol";
+// import {ISignatureValidator} from "../interfaces/ISignatureValidator.sol";
+import "../interfaces/modules/ISimpleExecutionModule.sol";
+
+contract SimpleExecutionModule is ISimpleExecutionModule, ReentrancyGuard {
+    /**
+     * @dev Safe (ex-Gnosis) style transaction
+     * @dev Allows to execute a transaction on SA.execute() internal method which opens up ability to do delegate calls
+     * @dev required to be called by a Smart Account
+     * @param _tx Smart Account transaction
+     */
+
+    // Should do validation using one of the validation modules..
+
+    function execTransaction(
+        Transaction memory _tx
+    ) public payable virtual nonReentrant returns (bool success) {
+        success = IExecFromModule(msg.sender).execTransactionFromModule(
+            _tx.to,
+            _tx.value,
+            _tx.data,
+            _tx.operation
+        );
+        if (!success) {
+            revert ExecutionFailed();
+        }
+    }
+}
--- a/test/bundler-integration/module/SimpleExecution.Module.specs.ts
+++ b/test/bundler-integration/module/SimpleExecution.Module.specs.ts
@@ -122,11 +122,6 @@ describe("ECDSA Registry Validation + Simple Execution Module (with Bundler):", 
     const tokensToMint = ethers.utils.parseEther("100");
     await mockToken.mint(bob.address, tokensToMint.toString());
 
-    await deployer.sendTransaction({
-      to: smartAccountOwner.address,
-      value: ethers.utils.parseEther("60"),
-    });
-
     return {
       entryPoint: entryPoint,
       smartAccountImplementation: await getSmartAccountImplementation(),
@@ -204,7 +199,19 @@ describe("ECDSA Registry Validation + Simple Execution Module (with Bundler):", 
         }
       );
 
-      await environment.sendUserOperation(userOp, entryPoint.address);
+      let thrownError: Error | null = null;
+
+      // let expectedError = new UserOperationSubmissionError()
+
+      // If bundler was supposed to throw from Validation.
+      try {
+        await environment.sendUserOperation(userOp, entryPoint.address);
+      } catch (e) {
+        thrownError = e as Error;
+        console.log("=================> error thrown here ==========>");
+      }
+
+      // expect(thrownError).to.deep.equal(expectedError);
 
       expect(await mockToken.balanceOf(bob.address)).to.equal(
         bobBalanceBefore.add(ethers.utils.parseEther("20"))
@@ -219,7 +226,10 @@ describe("ECDSA Registry Validation + Simple Execution Module (with Bundler):", 
       );
     });
 
-    /* it("Should not be able to execute if module is not enabled", async () => {
+    // TODO: negative test cases should be moved outside of bundler tests
+    // TODO: utils can be added to parse userOperationEvent and userOperationRevertReason
+    // Note: negative test case can have random sender calling this module method.
+    it("Should not be able to execute if module is not enabled", async () => {
       const {
         ecdsaRegistryModule,
         entryPoint,
@@ -230,12 +240,12 @@ describe("ECDSA Registry Validation + Simple Execution Module (with Bundler):", 
       } = await setupTests();
       // console.log(await userSA.getImplementation());
 
-    //   expect(await userSA.isModuleEnabled(delegateCallModule.address)).to.equal(
-    //     true
-    //   );
+      expect(await userSA.isModuleEnabled(delegateCallModule.address)).to.equal(
+        true
+      );
 
-      // const feeCollctorBalanceBefore = await mockToken.balanceOf(feeCollector);
-      // console.log("collector balance before ", feeCollctorBalanceBefore);
+      const feeCollctorBalanceBefore = await mockToken.balanceOf(feeCollector);
+      console.log("collector balance before ", feeCollctorBalanceBefore);
 
       // Checking enabled modules
       const returnedValueBefore = await userSA.getModulesPaginated(
@@ -244,11 +254,11 @@ describe("ECDSA Registry Validation + Simple Execution Module (with Bundler):", 
       );
       console.log("enabled modules before", returnedValueBefore);
 
-      // TODO: Review
+      // Accurate as modules are added in linked list in opposite order
       // Making a tx to disable a module
       const userOp1 = await makeEcdsaModuleUserOp(
         "disableModule",
-        [AddressOne, delegateCallModule.address],
+        [AddressOne, delegateCallModule.address], // in order to remove last added module prevModule would be Sentinel
         userSA.address,
         smartAccountOwner,
         entryPoint,
@@ -267,11 +277,11 @@ describe("ECDSA Registry Validation + Simple Execution Module (with Bundler):", 
       );
 
       // Logging modules after disabling
-    //   const returnedValue = await userSA.getModulesPaginated(
-    //     "0x0000000000000000000000000000000000000001",
-    //     10
-    //   );
-    //   console.log("enabled modules ", returnedValue); 
+      const returnedValue = await userSA.getModulesPaginated(
+        "0x0000000000000000000000000000000000000001",
+        10
+      );
+      console.log("enabled modules ", returnedValue);
 
       // Making the transaction using a module which should technically fail
       const wrapperCallData = mockWrapper.interface.encodeFunctionData(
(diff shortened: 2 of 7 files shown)


contracts/smart-account/modules/SimpleExecutionModule.sol
```diff
@@ -0,0 +1,31 @@
...
+     * @dev required to be called by a Smart Account
+     * @param _tx Smart Account transaction
+     */
+
+    // Should do validation using one of the validation modules..
+
+    function execTransaction(
+        Transaction memory _tx
+    ) public payable virtual nonReentrant returns (bool success) {
+        success = IExecFromModule(msg.sender).execTransactionFromModule(
```
filmakarov, line 21: I think we could in theory add some additional checks here regarding what's in the txn details. 
Or since this is the execution flow can add post checks after line 29, to check that, for example, implementation has not changed

contracts/smart-account/modules/SimpleExecutionModule.sol
filmakarov, line 1: What we've discussed with some customer, is that delegatecalls are safe until they do not mess with the SA storage.
In the case that we have discussed with the customer, the `delegatecall` to wrapper was needed to allow one operation to use the unpredictable outcome of another operation, which is not possible with `executeBatch`
So it doesn't write things to the SA storage.

So my main suggestion here is to introduce the check (maybe we need to introduce it to the `ModuleManager` as it's not easy to track SA storage access from Module) that if the `_tx.operation == 1`, the operation does not change SA storage. 
Thus it will allow for the "safe" delegatecalls, which allow batching some operations together in a sophisticated way but preserve SA storage consistency

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+      [delegateCallModule.address],
+      userSA.address,
+      smartAccountOwner,
+      entryPoint,
+      ecdsaRegistryModule.address
+    );
+
+    await entryPoint.handleOps([userOp1], bob.address);
+
+    console.log("delegate call module addr ", delegateCallModule.address);
```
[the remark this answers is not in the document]
livingrockrises: ok

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    /* const deploymentData = saFactory.interface.encodeFunctionData(
```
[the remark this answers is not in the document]
livingrockrises: ok

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+      );
+      console.log("enabled modules ", returnedValue);
+
+      const userSABalanceBefore = await mockToken.balanceOf(userSA.address);
+      const bobBalanceBefore = await mockToken.balanceOf(bob.address);
+      const feeCollctorBalanceBefore = await mockToken.balanceOf(feeCollector);
+
+      const wrapperCallData = mockWrapper.interface.encodeFunctionData(
+        "interact",
+        [mockToken.address, bob.address, ethers.utils.parseEther("30")]
```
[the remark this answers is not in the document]
livingrockrises: done

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+        ecdsaRegistryModule,
+        entryPoint,
+        userSA,
+        delegateCallModule,
+        mockWrapper,
+        mockToken,
+      } = await setupTests();
+      // console.log(await userSA.getImplementation());
+
+      expect(await userSA.isModuleEnabled(delegateCallModule.address)).to.equal(
```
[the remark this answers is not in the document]
livingrockrises: there is no harm in checking it

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+
+      // Review: how to check failures from bundler side
+      await environment.sendUserOperation(userOp1, entryPoint.address);
+
+      // Module should have been disabled correctly
+      expect(await userSA.isModuleEnabled(delegateCallModule.address)).to.equal(
+        false
+      );
+
+      // Logging modules after disabling
```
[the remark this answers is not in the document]
livingrockrises: done

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+      expect(await mockToken.balanceOf(userSA.address)).to.equal(
+        userSABalanceBefore.sub(ethers.utils.parseEther("30"))
+      );
+
+      expect(await mockToken.balanceOf(feeCollector)).to.equal(
+        feeCollctorBalanceBefore.add(ethers.utils.parseEther("10"))
+      );
+    });
+
+    // TODO: negative test cases should be moved outside of bundler tests
```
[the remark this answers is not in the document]
livingrockrises: ok will remove it from here

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+    // TODO: utils can be added to parse userOperationEvent and userOperationRevertReason
```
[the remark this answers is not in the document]
livingrockrises: yeah

test/bundler-integration/module/SimpleExecution.Module.specs.ts
```diff
@@ -0,0 +1,357 @@
...
+    // Note: negative test case can have random sender calling this module method.
```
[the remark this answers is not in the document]
livingrockrises: ok will cover this in different spec file in hardhat environment.

contracts/smart-account/modules/SimpleExecutionModule.sol
```diff
@@ -0,0 +1,31 @@
...
+
+contract SimpleExecutionModule is ISimpleExecutionModule, ReentrancyGuard {
+    /**
+     * @dev Safe (ex-Gnosis) style transaction
+     * @dev Allows to execute a transaction on SA.execute() internal method which opens up ability to do delegate calls
+     * @dev required to be called by a Smart Account
+     * @param _tx Smart Account transaction
+     */
+
+    // Should do validation using one of the validation modules..
```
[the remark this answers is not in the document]
livingrockrises: This is actual.
I mean it's not a must for non-4337 flow
but only SA can call this module and there would have been some level of authentication to dispatch tx from SA to here.

contracts/smart-account/modules/SimpleExecutionModule.sol
livingrockrises: hmm I am adding a note of this after line 29.

contracts/smart-account/modules/SimpleExecutionModule.sol
livingrockrises: why is it specific to executeBatch though?

didn't get this

> the delegatecall to wrapper was needed to allow one operation to use the unpredictable outcome of another operation, which is not possible with executeBatch
So it doesn't write things to the SA storage.

making a check in ModuleManager makes sense when operation is 1. What kind of checks you suggest written in solidity for storage layout?

test/bundler-integration/module/SimpleExecution.Module.specs.ts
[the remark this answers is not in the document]
filmakarov: no harm at all, I agree

it is just shown in the prev case, that the module is enabled successfully with the setupTests, so this check is not needed here.

contracts/smart-account/modules/SimpleExecutionModule.sol
filmakarov: No, it is not specific to executeBatch. 
What I meant is the use case, when let's say we want to:
1) swap token_A to token_B and then 
2) transfer 100% of the received token B.
It is not possible to achieve this with executeBatch, however, it is possible with delegatecalling the wrapper contract.
So this is the legit and secure use case for `delegatecalls`.
And it doesn't affect SA storage.
So we could introduce an optional check, that delegatecalls are allowed only if this call doesn't affect SA storage

filmakarov: > What kind of checks you suggest written in solidity for storage layout?

Thinking about it.
On the bundler side, it's easy. Just ensure no SSTORE opcodes are involved and all is good.

On a Solidity side of things, since the Smart Contract storage is huge addressed by random hashes, any of which can be used, I do not think that it is possible to somehow ensure the storage has not changed.
Ideally, we would need something like STORHASH (similar to EXTCODEHASH) opcode to achieve this

filmakarov: So maybe it makes sense to focus on achieving the same without `delegatecall`, or introducing a list of smart contracts which are allowed to be delegatecalled. Those contracts are checked not to change the state of a caller contract.

filmakarov: For the 1) the solution can be the following.

Build a module, which does the following:

pseudocode:
```
receivedAmount = SA.execTransactionFromModuleReturnData(...swap_TokenA_to_TokenB...);
SA.executeTransactionFromModule(...tokenB.transfer(receivedAmount)...)
```
Thus we do not need delegatecalls at all

test/bundler-integration/module/SimpleExecution.Module.specs.ts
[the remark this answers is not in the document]
livingrockrises: done. check the file in updated PR https://github.com/bcnmy/scw-contracts/pull/159

contracts/smart-account/modules/SimpleExecutionModule.sol
livingrockrises: > swap token_A to token_B and then
> transfer 100% of the received token B.

why this is only possible with delegate call?

livingrockrises: if wrapper is used for entire logic then delegate call is must. if batch entails that a b and c step should be done then call is good.

livingrockrises: > So maybe it makes sense to focus on:
> 
> 1. achieving the same without `delegatecall`,
> 
> or
> 
> 2. introducing a list of smart contracts which are allowed to be delegatecalled. Those contracts are checked not to change the state of a caller contract.



> > What kind of checks you suggest written in solidity for storage layout?
> 
> Thinking about it. On the bundler side, it's easy. Just ensure no SSTORE opcodes are involved and all is good.
> 
> On a Solidity side of things, since the Smart Contract storage is huge addressed by random hashes, any of which can be used, I do not think that it is possible to somehow ensure the storage has not changed. Ideally, we would need something like STORHASH (similar to EXTCODEHASH) opcode to achieve this

hmm there is no STORHASH opcode. like bundler does we'd have to debug trace and find accessed storage. not sure if there is a way to make storage snapshot before and after. @ankurdubey521 any thoughts?

livingrockrises: > For the 1) the solution can be the following.
> 
> Build a module, which does the following:
> 
> pseudocode:
> 
> ```
> receivedAmount = SA.execTransactionFromModuleReturnData(...swap_TokenA_to_TokenB...);
> SA.executeTransactionFromModule(...tokenB.transfer(receivedAmount)...)
> ```
> 
> Thus we do not need delegatecalls at all

Yes this is same as doing the same in executeBatch. 
some dapps want their custom wrapper for such logic that's when delegate calls become must.

filmakarov: > > swap token_A to token_B and then
> > transfer 100% of the received token B.
> 
> why this is only possible with delegate call?

Because the result of the swap operation is unpredictable. you know the result after the swap is executed (because of slippage).

filmakarov: > hmm there is no STORHASH opcode. like bundler does we'd have to debug trace and find accessed storage. not sure if there is a way to make storage snapshot before and after. @ankurdubey521 any thoughts?

yes, unfortunately there's no such opcode :( thus there's no way of enforcing 'no storage changes' rule with solidity/assembly

filmakarov: > Yes this is same as doing the same in executeBatch. some dapps want their custom wrapper for such logic that's when delegate calls become must.

Nope, it is not the same.
As with executeBatch you can not use the result of one (unpredictable) operation as an input for the next one.
With what I proposed, this is possible.

And afaic dApps want custom wrapper because they want exactly this : using the output of one operation as input for another.
If we suggest them the way of doing it w/o delegatecalls they could agree.
They will still have the custom wrapper, it will be just needed to enable it as module.
This would allow us to completely remove delegatecalls from `execTransactionFromModule`

filmakarov:
ii) This will help with protecting known slots, however, even with basic mapping we won't be able to check all the slots or it will be time-consuming. 
Also, writing something to the pseudo-random slot that is not checked can have a time-bomb effect

So, again, instead of doing this I suggest to better understand the need in delegatecalls and maybe we can avoid making `delegatecall` completely

PR #167, opened 2023-11-02
SMA-316: Remove dependence on via-ir for compilation

ankurdubey521:
# Summary

While we'll still use via-ir while deploying contracts, having contracts that do not require via-ir to be compiled enables the coverage-tracking plugins to work properly.
I've made very simple modifications that should in theory not affect the gas consumed by these contracts. But still to be on the safe side, what would be the best way to measure gas (before and after) for the updated modules? @filmakarov 

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [x] Other

# Checklist

### DIFF
--- a/contracts/smart-account/modules/AccountRecoveryModule.sol
+++ b/contracts/smart-account/modules/AccountRecoveryModule.sol
@@ -150,35 +150,37 @@ contract AccountRecoveryModule is
         address lastGuardianAddress;
         address currentGuardianAddress;
         bytes memory currentGuardianSig;
-        uint48 validAfter;
-        uint48 validUntil;
         uint48 latestValidAfter;
         uint48 earliestValidUntil = type(uint48).max;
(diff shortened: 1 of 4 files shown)


Aboudjem (approved):

livingrockrises (commented):
looks good to me using new block scopes. 
can you remove viaIR from hardhat config as well?

ankurdubey521:
@livingrockrises we should still keep via-ir enabled for deploying contracts to production, these changes are just to make sure when coverage is run (which cannot use via-ir) the contracts still compile.

livingrockrises (approved):

ankurdubey521:
@filmakarov it would be great to get your review as well, since you're the primary author of these modules.

PR #168, opened 2023-11-06
SMA-27 Account Recovery Module: fixes upon audit

filmakarov:
# Summary
- Added a check for the outer method selector for the userOp.callData to avoid call to other methods rather than default executor
- Now guardians sign controlMessage + SA.address. This is even more secure in terms of revealing guardians address across SAs.
- Removed unneeded conditional

TODO: technically `validAfter = request.timestamp + securityDelay` may overflow as both `validAfter` and `securityDelay` are `uint48`s. Making `securityDelay` an uint24 should solve issue however it will limit delay to 194 days (which is enough I think).
Waiting confirmation from auditors on this matter.

Related Issue: # [SMA-27](https://linear.app/biconomy/issue/SMA-27/account-recovery-zellic)

## Change Type

- [x] Bug Fix

# Checklist


contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -34,11 +34,13 @@ contract AccountRecoveryModule is
...
-    // keccak256(abi.encodePacked("ACCOUNT RECOVERY GUARDIAN SECURE MESSAGE"))
-    bytes32 public constant CONTROL_HASH =
-        0x1c3074d7ce4e35b6f0a67e3db0f31ff117ca8f0a1853e86a95837fea35af828b;
-    bytes32 public immutable controlHashEthSigned =
-        CONTROL_HASH.toEthSignedMessageHash();
+    // execute(address,uint256,bytes)
+    bytes4 public constant EXECUTE_SELECTOR = 0xb61d27f6;
+    // execute_ncC(address,uint256,bytes)
+    bytes4 public constant EXECUTE_OPTIMIZED_SELECTOR = 0x0000189a;
+    // Hash to be signed by guardians to make a guardianId
```
Aboudjem, line 41: wouldn't hardcoding these values make this module incompatible with other SmartAccounts? Should we stick to this or make these variables "immutable" and set them in the constructor?

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -167,9 +169,9 @@ contract AccountRecoveryModule is
             currentGuardianSig = moduleSignature[(2 * i + 1) * 65:(2 * i + 2) *
                 65];
 
-            currentGuardianAddress = controlHashEthSigned.recover(
-                currentGuardianSig
-            );
+            currentGuardianAddress = keccak256(
+                abi.encodePacked(CONTROL_MESSAGE, userOp.sender)
+            ).toEthSignedMessageHash().recover(currentGuardianSig);
```
Aboudjem, line 174: see above msg

contracts/smart-account/modules/AccountRecoveryModule.sol
```diff
@@ -34,11 +34,13 @@ contract AccountRecoveryModule is
...
-        0x1c3074d7ce4e35b6f0a67e3db0f31ff117ca8f0a1853e86a95837fea35af828b;
-    bytes32 public immutable controlHashEthSigned =
-        CONTROL_HASH.toEthSignedMessageHash();
+    // execute(address,uint256,bytes)
+    bytes4 public constant EXECUTE_SELECTOR = 0xb61d27f6;
+    // execute_ncC(address,uint256,bytes)
+    bytes4 public constant EXECUTE_OPTIMIZED_SELECTOR = 0x0000189a;
+    // Hash to be signed by guardians to make a guardianId
+    string public constant CONTROL_MESSAGE =
+        "ACCOUNT RECOVERY GUARDIAN SECURE MESSAGE";
```
Aboudjem, line 43: Making the message fit to  `32bytes` saves gas on deployment (~2200 gas) and on `abi.encodePacked` (~80-100 gas)

Aboudjem (changes requested):

livingrockrises:
resolve conflicts

contracts/smart-account/interfaces/IAccountRecoveryModule.sol
```diff
@@ -271,7 +271,7 @@ interface IAccountRecoveryModule {
      * Should be called by the Smart Account
      * @param newSecurityDelay new security delay
      */
-    function setSecurityDelay(uint48 newSecurityDelay) external;
+    function setSecurityDelay(uint24 newSecurityDelay) external;
```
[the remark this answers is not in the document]
filmakarov: The reason for modification here was to avoid potential overflow when the `securityDelay` is added to the `validAfter` param, and the `validAfter` is `uint48` as per 4337

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: Fair point. I can do this. However, in any case any SA with other selectors will require to redeploy the module with own parameters then.

I think this issue won't be an issue, when SA execution selectors will be standardized (via 6900 or some other eip)

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: done

contracts/smart-account/modules/AccountRecoveryModule.sol
filmakarov: +

filmakarov:
all comments addressed

Aboudjem (approved):
🚀

PR #169, opened 2023-11-06
Feat/address resolver

livingrockrises:
# Summary

resolves V1 and V2 smart account address for given eoa address and range of deployment indexes. for V2 it assumes default auth module is ECDSA module

latest contract: https://mumbai.polygonscan.com/address/0x46A82fB84089c0cFfE9685ACe7Be821cfd7f441d#readContract 

Related Issue: # (issue number)

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

# Additional Information

# Branch Naming

### DIFF
--- a/contracts/smart-account/libs/AddressResolver.sol
+++ b/contracts/smart-account/libs/AddressResolver.sol
@@ -18,10 +18,21 @@ interface IEcdsaRegistryModule {
     function initForSmartAccount(address eoaOwner) external returns (address);
 }
 
+interface ISmartAccount {
+    function getImplementation()
+        external
+        view
+        returns (address _implementation);
+
+    function VERSION() external view returns (string memory);
+}
+
 contract AddressResolver {
-    struct SmartAccountResult {
+     struct SmartAccountResult {
         address accountAddress;
         address factoryAddress;
+        address currentImplementation;
+        string currentVersion;
         string factoryVersion;
     }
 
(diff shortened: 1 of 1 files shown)


contracts/smart-account/libs/AddressResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+        0x00006B7e42e01957dA540Dc6a8F7C30c4D816af5;
+    // marked for deletion
+    address public constant SA_V2_IMPLEMENTATION =
+        0x0000002512019Dafb59528B82CB92D3c5D2423aC;
+    address public constant ECDSA_REGISTRY_MODULE_ADDRESS =
+        0x0000001c5b32F37F5beA87BDD5374eB2aC54eA8e;
+
+    // Todo Add natspec
+    // Review could pass factory and module address in params
+    function resolveEOAtoAccountAddresses(
```
ankurdubey521, line 55: factory, module and setupdata should be passed as parameters here so that we can use them with other modules

contracts/smart-account/libs/AddressResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+        uint256 index
+    ) external view returns (address _account);
+
+    function getAddressForCounterFactualAccount(
+        address _owner,
+        uint256 _index
+    ) external view returns (address _account);
+}
+
+interface IEcdsaRegistryModule {
```
ankurdubey521, line 17: should import relevant interfaces and not re-define here

contracts/smart-account/libs/AddressResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+    function initForSmartAccount(address eoaOwner) external returns (address);
+}
+
+interface ISmartAccount {
```
ankurdubey521, line 21: same here

contracts/smart-account/libs/AddressResolver.sol
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+interface IAddressResolver {
```
ankurdubey521, line 4: same here

contracts/smart-account/libs/AddressResolver.sol
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
```
ankurdubey521, line 2: use solidity ^0.8.20

ankurdubey521:
Also, since this is not a library it should not be in the libs folder

contracts/smart-account/libs/AddressResolver.sol
livingrockrises: I feel this is very specific to Biconomy Factories. and only for V1 and V2

like they must have to pass V1 factory address
and they must have to pass V2 factory address
which are fixed.

module address and module setup data can be passed in the method yes. but they will also only work with V2.

hence it's difficult to extend this contract to work with wider use cases and currently specific case (for biconomy, for v1 and v2 only) is enough.

livingrockrises:
> Can we add unit tests for this?

can add a basic test using V2 factory only and auth modules like ecdsa and multi chain ecdsa.

contracts/smart-account/libs/AddressResolver.sol
livingrockrises: got it.

contracts/smart-account/libs/AddressResolver.sol
livingrockrises: once your PR is merged would be better to do this just for this contract.

contracts/smart-account/libs/AddressResolver.sol
livingrockrises: done

contracts/smart-account/libs/AddressResolver.sol
livingrockrises: done

livingrockrises:
updated PR
https://github.com/bcnmy/scw-contracts/pull/171

contracts/smart-account/libs/AddressResolver.sol
ankurdubey521: got it

PR #175, opened 2023-11-28
SMA-146 Only Module Removal Check

filmakarov:
# Summary

Added the check to not allow removing the only enabled module

Related Issue: [#SMA-146](https://linear.app/biconomy/issue/SMA-146/implement-check-to-prevent-removal-of-last-module)

## Change Type

- [ ] Other

# Checklist

### DIFF
--- a/contracts/smart-account/interfaces/base/IModuleManager.sol
+++ b/contracts/smart-account/interfaces/base/IModuleManager.sol
@@ -77,6 +77,12 @@ interface IModuleManager {
     );
 
     /**
+     * @notice Throws when trying to remove the only enabled module
+     * @param module Module address provided
+     */
+    error CanNotDisableOnlyModule(address module);
+
+    /**
      * @dev Adds a module to the allowlist.
      * @notice This SHOULD only be done via userOp or a selfcall.
      */
--- a/contracts/smart-account/base/ModuleManager.sol
+++ b/contracts/smart-account/base/ModuleManager.sol
@@ -217,6 +217,8 @@ abstract contract ModuleManager is SelfAuthorized, Executor, IModuleManager {
         if (module == address(0) || module == SENTINEL_MODULES) {
             revert ModuleCannotBeZeroOrSentinel(module);
         }
+        if (_modules[module] == SENTINEL_MODULES && _modules[SENTINEL_MODULES] == module)
+            revert CanNotDisableOnlyModule(module);
         if (_modules[prevModule] != module) {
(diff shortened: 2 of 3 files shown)


ankurdubey521 (approved):

livingrockrises:
nice

livingrockrises:
I think there is duplicate task on this can you confirm
https://linear.app/biconomy/issue/SMA-160/security-fix-for-module-removal

livingrockrises (approved):

contracts/smart-account/interfaces/base/IModuleManager.sol
```diff
@@ -76,6 +76,12 @@ interface IModuleManager {
         uint256 operationLength
     );
 
+    /**
+     * @notice Throws when trying to remove the only enabled module
+     * @param module Module address provided
+     */
+    error CanNotDisableOnlyModule(address module);
```
Aboudjem, line 83: I'll suggest `Cannot` instead of `CanNot` since we already have  `ModuleCannotBeZeroOrSentinel`

contracts/smart-account/interfaces/base/IModuleManager.sol
```diff
@@ -76,6 +76,12 @@ interface IModuleManager {
         uint256 operationLength
     );
 
+    /**
+     * @notice Throws when trying to remove the only enabled module
+     * @param module Module address provided
```
Aboudjem, line 81: Can you please add a note here or in `_disableModule` or both, stating that this does not cover the case where the last module is an "execution" module? If the last module is an execution module, the SA will make it unrecoverable. It's important that everyone understands the implications of removing a module.

We should really think about an identifier for the modules.

Aboudjem (changes requested):

filmakarov:
> I think there is duplicate task on this can you confirm https://linear.app/biconomy/issue/SMA-160/security-fix-for-module-removal

yes, this is duplicate

contracts/smart-account/interfaces/base/IModuleManager.sol
filmakarov: I made this addition. 
We can introduce self-declaration for the validation modules via ERC165, however, this doesn't solve for cases when malicious module self declares being validation module, while it is not

contracts/smart-account/interfaces/base/IModuleManager.sol
filmakarov: made both `CanNot`

PR #178, opened 2023-11-30
SMA-338: Verify returndatasize from setupAndEnableModule()

ankurdubey521:
# Summary
Fix informational issue raised by Zellic in the Audit for Security Policies

## Change Type

- [x] Bug Fix
- [ ] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

### DIFF
--- a/contracts/smart-account/modules/SecurityPolicyManagerPlugin.sol
+++ b/contracts/smart-account/modules/SecurityPolicyManagerPlugin.sol
@@ -87,8 +87,10 @@ contract SecurityPolicyManagerPlugin is ISecurityPolicyManagerPlugin {
 
             switch success
             case 0x1 {
-                moduleInstallationSuccess := mload(ptr)
-                module := mload(add(ptr, 0x60))
+                if eq(size, 0x80) {
+                    moduleInstallationSuccess := mload(ptr)
+                    module := mload(add(ptr, 0x60))
+                }
             }
             case 0x0 {
                 revert(ptr, size)


AmanRaj1608 (approved):

filmakarov (approved):

contracts/smart-account/modules/SecurityPolicyManagerPlugin.sol
```diff
@@ -87,8 +87,10 @@ contract SecurityPolicyManagerPlugin is ISecurityPolicyManagerPlugin {
 
             switch success
             case 0x1 {
-                moduleInstallationSuccess := mload(ptr)
-                module := mload(add(ptr, 0x60))
+                if eq(size, 0x80) {
```
livingrockrises, line 90: let success := call(
                gas(),
                caller(),
                0,
                savePtr,
                sub(ptr, savePtr),
                0,
                0 **// provide expected return size here**
            )
            
            Should't this be enough?

ankurdubey521: So the issue here is that in a situation where the setupContract doesn't return the module address after setupCalldata is executed, then the returndatasize of this call would just be 0x20 (success).
If it does return then the length of 0x80 (success + offset + length + module address).
So even if we hard-coded 0x80 in the call, we'll still have to manually check if 0x80 bytes were actually returned by calling returndatasize

PR #193, opened 2024-01-20
Fix/excess override on is valid signature unsafe

filmakarov:
# Summary

AS per Chirag request

### DIFF
--- a/contracts/smart-account/modules/AccountRecoveryModule.sol
+++ b/contracts/smart-account/modules/AccountRecoveryModule.sol
@@ -489,7 +489,7 @@ contract AccountRecoveryModule is
     function isValidSignature(
         bytes32,
         bytes memory
-    ) public view virtual override returns (bytes4) {
+    ) public view virtual returns (bytes4) {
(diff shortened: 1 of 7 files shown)


Aboudjem (approved):

livingrockrises:
umm looks like all of these functions come from corresponding interfaces and not base contracts we are deriving from. and all interface functions are implicitly virtual. so override is merely clarifying and it's fine

livingrockrises:
> Using override explicitly indicates that you are implementing a function from an interface or a base contract. While it is not strictly necessary, it can make the code more readable, especially in larger codebases where it might not be immediately clear that a particular function is part of an interface

:/ my bad I didn't see they are from interfaces.

livingrockrises:
I actually copied AccountRecoveryModule from develop branch to v2-deployments-with-config branch but now I realise interfaces in later branch don't have these methods and hence my override didn't work which led me into this thinking and initiate conversation / suggestion with you.

(My opinion close PR with this comment)

